PR #14, opened 2021-05-06
Fix writing to plugin endpoint.

### LINKED ISSUE #13: '/plugins/' missing from URL on write to plugin endpoint
Changing custom plugin model via API produces following error:
```
pynautobot.core.query.RequestError: The requested url: https://$SERVER/api/ip_health_check/checks/ec131d77-41ec-4e30-a6ed-180bb6032e5b/ could not be found.
```
The issue is in URl missing `plugins` part in the request.

It is easily reproducible by having plugin with custom model n the DB. Let's say you have plugin `ip_health_check` that stores `result: bool` and link it to IP. This plugin has registered route with view on `checks`. The correct URL is `$SERVER/api/plugins/ip_health_check/checks/`.
The following code should change the result to False and save it to DB, but result is above error.
```
check = nb.plugins.ip_health_check.checks.filter(ip_address="10.24.50.128")[0]
check.result=False
check.save()
```

Thetacz:
Fixes #13. This change keeps `plugins/` as part of the app path and as results corrects the URL.

Integration tests are not working on my local setup, but everything else passed.


jmcgill298:
@Thetacz the URL resolution is currently flawed in that it works because the maintainers follow a prescribed pattern. This pattern is not enforced, and so plugins can deviate from the pattern established in the core Apps/Models. If I create an API endpoint using `""`, then pynautobot will return an App instance for what is actually an Endpoint. If I follow the pattern of Nautobot where the URL is App/Model, then I am able to get an endpoint correctly, but there is no conversion of `_` to `-` in the URL (nor should there be, as `_` is valid). I believe the correct approach is to inspect the API for valid plugin apps/models/endpoints instead of assuming a certain pattern.

In the meantime, I need to look further at this solution, but I would that that creating an instance of [PluginsApp](https://github.com/nautobot/pynautobot/blob/4e1033eb952d7eb3fbc699be8e103f594a174cd7/pynautobot/core/app.py#L125) should solve for the current use case. Have you tried this solution already?

Thetacz:
Hi @jmcgill298, sorry it took me so long to respond. I'm sorry but I did not have time to think of more rigid solution than the one I offered. 
> @Thetacz the URL resolution is currently flawed in that it works because the maintainers follow a prescribed pattern. This pattern is not enforced, and so plugins can deviate from the pattern established in the core Apps/Models. If I create an API endpoint using "", then pynautobot will return an App instance for what is actually an Endpoint. If I follow the pattern of Nautobot where the URL is App/Model, then I am able to get an endpoint correctly, but there is no conversion of _ to - in the URL (nor should there be, as _ is valid). I believe the correct approach is to inspect the API for valid plugin apps/models/endpoints instead of assuming a certain pattern.

AFAIK with my change the documented way is working and undocumented not, without it following documentation leads to error. Seems to me like and improvement. But I agree the ideal solution would cover all cases.

> In the meantime, I need to look further at this solution, but I would that that creating an instance of [PluginsApp](https://github.com/nautobot/pynautobot/blob/4e1033eb952d7eb3fbc699be8e103f594a174cd7/pynautobot/core/app.py#L125) should solve for the current use case. Have you tried this solution already?

The way [PluginsApp](https://github.com/nautobot/pynautobot/blob/4e1033eb952d7eb3fbc699be8e103f594a174cd7/pynautobot/core/app.py#L125)  works quite confuses me so I have not tried digging around that part. Can you please elaborate on the proposed solution? Thank you

pynautobot/core/response.py
```diff
@@ -280,8 +280,9 @@ def _endpoint_from_url(self, url):
             url_path = url_path[len(extra_path) :]
         split_url_path = url_path.split("/")
         if split_url_path[2] == "plugins":
-            # Skip plugins from the path
-            app, name = split_url_path[3:5]
+            # Keep plugins in app path
+            app = "/".join(split_url_path[2:4])
+            name = split_url_path[4]
```
jmcgill298, line 285:
```suggestion
            app, name = split_url_path[3:5]
            return getattr(pynautobot.core.app.PluginsApp(self.api, app), name)
```
```suggestion
            # Keep plugins in app path
            app = "/".join(split_url_path[2:4])
            name = split_url_path[4]
```

jmcgill298, line 285:
```suggestion
            # Keep plugins in app path
            app, name = split_url_path[3:5]
            return getattr(pynautobot.core.app.PluginsApp(self.api, app), name)
```

jmcgill298: I'm not sure if this works, but this is what I would expect

Thetacz: Sadly that does not work. AFAICT `PluginsApp` is written somehow strangle as it does not extend original `App` and just adds `plugins/` to the url, but instead it is completely new object, so this way you never get to `App.__getattr__`. And it has different `__init__`. The result is your suggestion cannot work as `PluginsApp` does no accept `app`. If I try to pass the whole string there as `app/name` in place of `name`, then `Endpoint.url` does not get set as string so it crashes in `Endpoint.normalize_url` function.
So to fix #13 it either requires my easy-but-ugly fix or refactor `PluginsApp`...

jmcgill298:
@Thetacz I believe if you pull in the updates from the develop branch then the tests will pass now

jmcgill298 (approved):
Thanks @Thetacz

PR #17, opened 2021-05-28
fix #16 Print virtual_chassis

### LINKED ISSUE #16: Error when printing Virtual-Chassis object
there is an error when printing a Virtual Chassis Record

```
print(record)
  File "<string>", line 1, in <module>
  File "/home/ansible/devel/nautobot-transfer/lib/python3.8/site-packages/pynautobot/models/dcim.py", line 151, in __str__
    return self.master.display_name
AttributeError: 'NoneType' object has no attribute 'display_name'
```

I could see that the Nautobot API returns the field `name` instead of `display_name`

I'll filing a pull request with changes `pynautobot/pynautobot/models/dcim.py` from

```python
class VirtualChassis(Record):
    def __str__(self):
        return self.master.display_name
```

to

```python
class VirtualChassis(Record):
    def __str__(self):
        if self.master is None:
            return self.name
        else:
            return self.master.display_name
```

jifox:
I've implemented that the current virtual_chassis name will be displayed if the master is None.

I have no example for that assigned master, so i could not determine if the original statement if correct with "display_name" or this should also be changed to "name"


jifox:
I'm not sure why the test failed. I'm able to successfully run my script with the suggested changes.

image

jmcgill298:
Thanks for the submission @jifox 

If you rebase off of develop, then the tests should pass now.

A couple of comments though:

1. Why are you proposing to use `self.name` instead of `self.display_name`?
2. I think the conditional ordering should be reversed `if self.master is not None...else...`; it is likely that self.master is not None more often than it is None, so that condition should take precedence.

jifox:
@jmcgill298 After a while I've implemented the suggested changes now

jvanderaa (approved):

Issue #44, opened 2022-04-26
pynautobot 1.1.0 Errors when trying to insert a cable record


gneville-ot:
I'm using pynautobot 1.1.0 with nautobot 1.3.1 and trying to insert a new cable entry but it errors. The same works with pynautobot 1.0.4 (with nautobot 1.3.1)

Example code:

```
cable_record = {'termination_a_type': 'dcim.interface', 'termination_a_id': source.id, 'termination_b_type': 'dcim.interface', 'termination_b_id': target.id, 'status': 'connected'}

nb.dcim.cables.create(cable_record)
```

This errors with:

```
  File "/home/appuser/.local/lib/python3.8/site-packages/pynautobot/core/endpoint.py", line 293, in create
    req = Request(
  File "/home/appuser/.local/lib/python3.8/site-packages/pynautobot/core/query.py", line 339, in post
    return self._make_call(verb="post", data=data)
  File "/home/appuser/.local/lib/python3.8/site-packages/pynautobot/core/query.py", line 241, in _make_call
    raise RequestError(req)
pynautobot.core.query.RequestError: The request failed with code 406 Not Acceptable: {'detail': 'Invalid version in "Accept" header. Supported versions are 1.2, 1.3'}
```

I'm getting up pynautobot with the following:

```
nautobot = nb_api(url=nb_url, token=nb_token)
```

I have tried to add `api_version="1.3"` but it fails with the same.

Doing some debugging if I edit `pynautobot/core/query.py`and print the headers in _make_call after `if self.api_version:` I see the following:

```
{'Content-Type': 'application/json;', 'authorization': 'Token <token>', 'accept': "application/json; version={'termination_a_type': 'dcim.interface', 'termination_a_id': 'cd80ed0b-96c9-49ca-afa2-78d79a18cfaf', 'termination_b_type': 'dcim.interface', 'termination_b_id': 'e6651be5-16bc-4828-a173-5559ff482309', 'status': 'connected'}"}
```

So it looks like the headers are getting mangled with the post data. 

If I revert to 1.0.4 everything is fine, the headers are normal:

```
{'Content-Type': 'application/json;', 'authorization': 'Token <token>'}
```

FragmentedPacket:
Sorry about all the regressions. We're planning to refactor and increase test coverage but will try and get a fix out today or tomorrow.

PR #49, opened 2022-04-27
Testing improvements

### LINKED ISSUE #46: Tests do not consistently assert_called_with to confirm correct HTTP requests
Observed while troubleshooting #44 and trying to determine why this breakage wasn't detected in testing. For example:

```python
    @patch(
        "requests.sessions.Session.post", return_value=Response(fixture="dcim/device.json"),
    )
    def test_create(self, *_):
        data = {
            "name": "test-device",
            "site": 1,
            "device_type": 1,
            "device_role": 1,
        }
        ret = nb.devices.create(**data)
        self.assertTrue(ret)
```

At no point does this test confirm that `requests.post` is called *correctly* or *as expected*. **Any** call to `requests.post` will return the fixture response. This is a significant test gap.

glennmatthews:
- Add a couple of missing `assert_called_with` checks in test_dcim.DeviceTestCase; one of these updated tests now fails (as it should!) on account of #44. Fixes #46.
- Instead of having a duplicate `Generic.Tests` class in each app test file, deduplicate this class to the base `tests/__init__.py` file and re-use it as appropriate.
  - I also added a number of convenient helper properties to this class to reduce repetition of code a bit further.
- The test fixture files are still basically unchanged from pynetbox, including the use of integer rather than UUID PKs. I've made spot fixes to some of the fixture data to replace some PKs with a UUID, but these definitely need more love in the future to make them realistically reflect Nautobot's REST API.
- Fix an error in `tests/integration/conftest.py` on re-runs -- now that we're pinning to a specific commit of the devicetypes library, `git pull` would fail as we have checked out a commit, not a branch.
- Clean up some various `assertTrue` in the tests to more accurate `assertIsInstance` and/or `assertEqual` as appropriate


FragmentedPacket (approved):

jmcgill298 (approved):
Thanks Glenn!

Issue #53, opened 2022-05-03
AttributeError being raised for prefix object when trying to retrieve tags


thatfatguypat:
In my org we add tags to network prefixes as an identifier, and use those tags as a filter to find prefixes.  We then perform different operations based on the combination of tags assigned to the prefixes.  I can still use the filter on a get(), which provides me a list of prefixes with the tag, but I can no longer retrieve all tags assigned to the resulting prefixes by accessing the `tags` attribute of the prefixes.  

The code below would previously return a list of tags for `prefix_obj.tags` in v1.0.4.  However, in v1.1.1 the same code raises an AttributeError for the 'prefix' attribute being missing.

```
>>> import pynautobot
>>> url="https://my.nautobot.net"
>>> token='1234mocktoken5678'
>>> nautobot=pynautobot.api(url=url,token=token)
>>> name='prefix-tag-1'
>>> prefix = '10.0.0.0/26'
>>> network = prefix.split("/")[0]
>>> mask = int(prefix.split("/")[1])
>>> vrf = 'my-vrf-uuid-here'
>>> prefix_obj = nautobot.ipam.prefixes.get(q=network, mask_length=mask, vrf_id=vrf)
>>> type(prefix_obj)
<class 'pynautobot.models.ipam.Prefixes'>
>>> prefix_obj.tags

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/pynautobot/core/response.py", line 216, in __repr__
    return str(self)
  File "/usr/local/lib/python3.7/site-packages/pynautobot/models/ipam.py", line 29, in __str__
    return str(self.prefix)
  File "/usr/local/lib/python3.7/site-packages/pynautobot/core/response.py", line 197, in __getattr__
    raise AttributeError('object has no attribute "{}"'.format(k))
AttributeError: object has no attribute "prefix"
```
I noticed this change in [response.py](https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/response.py#L174-L176) from #41 which redefined `self.default_ret = self.endoint.return_obj`.  Reverting this appears to resolve my issue, but I don't have enough context as to the intent of the original change so I am not sure if that is the correct course of action.

serialize() on the prefix object, with some data redacted, shows that 'prefix' attribute exists, contrary
to what is indicated by the AttributeError:
```
>>> prefix_obj.serialize()
{'id': '<object-uuid-here>', 
[...]

PR #54, opened 2022-05-10
ipam.Prefixes.__str__ should return display attribute when prefixes attribute is missing

thatfatguypat:
Addresses https://github.com/nautobot/pynautobot/issues/53

### DIFF
--- a/pynautobot/models/ipam.py
+++ b/pynautobot/models/ipam.py
@@ -26,7 +26,7 @@ class IpAddresses(Record):
 
 class Prefixes(Record):
     def __str__(self):
-        ret = self.prefix if hasattr(self, 'prefix') else self.display
+        ret = self.prefix if hasattr(self, "prefix") else self.display
         return str(ret)
 
     @property
--- a/pynautobot/models/ipam.py
+++ b/pynautobot/models/ipam.py
@@ -26,7 +26,8 @@ class IpAddresses(Record):
 
 class Prefixes(Record):
     def __str__(self):
-        return str(self.prefix)
+        ret = self.prefix if hasattr(self, 'prefix') else self.display
+        return str(ret)
 
     @property
     def available_ips(self):


glennmatthews (commented):
It would *probably* be okay to just return `self.display` in all cases, but I'm OK with this more cautious approach.

Is `prefixes.tags` the only endpoint that shows this issue, or is this a general problem with any `Record` subclass that defines a custom `__str__` method - for example, is a similar fix needed on `IpAddresses`, above?

glennmatthews:
See my comment on #53 - I still think this change is valid in and of itself, but I don't believe it addresses the actual underlying issue, which may be a regression introduced by #41.

jmcgill298:
The issue seems to be what class is being used to generate the value(s) for attributes that are instances of another Nautobot object (Record classes in pynautobot). I put a different type of patch together in #55

thatfatguypat:
> Is prefixes.tags the only endpoint that shows this issue, or is this a general problem with any Record subclass that defines a custom __str__ method - for example, is a similar fix needed on IpAddresses, above?

As of yet, I've only been able to repro the issue with prefixes.tags

> I still think this change is valid in and of itself, but I don't believe it addresses the actual underlying issue, which may be a regression introduced by https://github.com/nautobot/pynautobot/pull/41.

@glennmatthews that's a fair assessment.  It very well may mean that I'm addressing a symptom, rather than the cause.  If that is the case, shall I close this out in preference for #53 or #55?

glennmatthews:
> > I still think this change is valid in and of itself, but I don't believe it addresses the actual underlying issue, which may be a regression introduced by #41.
> 
> @glennmatthews that's a fair assessment. It very well may mean that I'm addressing a symptom, rather than the cause. If that is the case, shall I close this out in preference for #53 or #55?

Let's keep this open for now - we're having some team discussions and investigation around #53 and #41 to figure out the best way forward.

thatfatguypat:
Closing this now that #57 has been merged and addresses Issue #53

PR #55, opened 2022-05-11
Possible Fix for #53

jmcgill298:
Overwrite the default return class for tag records, and extensible to include other record types


jmcgill298:
@glennmatthews @FragmentedPacket This seems like maybe a little bit of a hack, but I'm not sure why the `default_ret` attribute was overwritten in #41 . I did test out the scenario in #53 and it did work.

Issue #32, opened 2022-03-04
Add method to easily run a Job from pynautobot


dgarros:
it would be useful to have a method to easily start the execution of a Job with pynautobot

Something like that
```python
>>> import pynautobot
>>> nautobot = pynautobot.api( url="https://demo.nautobot.com/", token="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
>>>
>>> job = nautobot.extras.jobs.run(class_path="xxx", commit=False, data=None, schedule=None)
```

Also, maybe a separate request but it would be nice to have a way to execute a Job and wait for it to finish
Something like that
```python
>>> import pynautobot
>>> nautobot = pynautobot.api( url="https://demo.nautobot.com/", token="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
>>> job_result = nautobot.extras.jobs.run_and_wait(class_path="xxx", commit=False)
```

Renrut5:
@jvanderaa @jeffkala  Would this be considered completed with #56 merged?

PR #56, opened 2022-05-12
add run method to be able to execute jobs

jeffkala:
partially implements #32

### DIFF
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -397,7 +397,7 @@ class Endpoint(object):
     def run(self, *args, api_version=None, **kwargs):
         r"""Runs a job based on the class_path provided to the job.
 
-        Takes a kwarg of `class_path` to specify the job that should be run.
+        Takes a kwarg of `class_path` or `job_id` to specify the job that should be run.
 
         :arg str,optional \*args: Freeform search string that's
             accepted on given endpoint.
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -18,7 +18,7 @@ This file has been modified by NetworktoCode, LLC.
 from pynautobot.core.query import Request, RequestError
 from pynautobot.core.response import Record
 
-RESERVED_KWARGS = ("id", "pk", "limit", "offset")
+RESERVED_KWARGS = ("pk", "limit", "offset")
 
 
 def response_loader(req, return_obj, endpoint):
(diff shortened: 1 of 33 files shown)


pynautobot/core/endpoint.py
```diff
@@ -394,6 +394,56 @@ def count(self, *args, api_version=None, **kwargs):
...
+                commit=True,
+            )
+
+        >>>
+        """
+
+        if "jobs" not in self.url:
+            raise ValueError('The "run" method can only be executed against the "jobs" app.')
+
+        api_version = api_version or self.api.api_version or self.api.version
```
jvanderaa, line 427: Would this make sense in case none of those were found?
```suggestion
        api_version = api_version or self.api.api_version or self.api.version or None
        
        if api_version is None:
            raise Error
```

jeffkala: self.api.version is a class property that is returned from querying nautobot itself for the version.  Don't see a way it can be None.

pynautobot/core/endpoint.py
```diff
@@ -394,6 +394,56 @@ def count(self, *args, api_version=None, **kwargs):
 
         return ret.get_count()
 
+    def run(self, *args, api_version=None, **kwargs):
```
jmcgill298, line 397: I don't think this should be part of the endpoint class, but should be specific to a job class implementation

pynautobot/core/endpoint.py
```diff
@@ -394,6 +394,56 @@ def count(self, *args, api_version=None, **kwargs):
...
+                )
+            job_run_url = f"{self.url}/{kwargs['class_path']}/run/"
+        else:
+            if not kwargs.get("job_id"):
+                raise ValueError('Keyword Argument "job_id" is required to run a job in Nautobot APIv1.3 and newer.')
+            job_run_url = f"{self.url}/{kwargs['job_id']}/run/"
+
+        req = Request(
+            base=job_run_url, token=self.token, http_session=self.api.http_session, api_version=api_version,
+        ).post(args[0] if args else kwargs)
```
jmcgill298, line 443: how are jobs followed up on?

jmcgill298 (commented):
We will need to add documentation on how to start and view jobs once the implementation is finalized

pynautobot/core/endpoint.py
jvanderaa: @jeffkala

pynautobot/core/endpoint.py
jvanderaa: @jeffkala

pynautobot/core/endpoint.py
jvanderaa: @jeffkala adding this to your plate that is already full.

jvanderaa: We should be looking to return the JobResult ID to be able to follow up on if we do not have that.

jvanderaa (approved):

jvanderaa (approved):

PR #42, opened 2022-04-21
Changed __str__ to use the display field as it is on all models.

FragmentedPacket:
Fixes #34 

All objects from Nautobot have the display field and if not, it will make it a blank string.

PR #57, opened 2022-05-16
Rollback recent changes but accounts for several workarounds that were in place

### LINKED ISSUE #53: AttributeError being raised for prefix object when trying to retrieve tags

FragmentedPacket:
Changes:
 - This reverts the changes in #41 as they weren't the appropriate fix and caused more workarounds to have to be put in place.
 - Reverting #41 and the change for the base `__str__` method on a `Record` object to **self.display** that fixes the nat_inside/nat_outside not having the correct attributes available is within #42.
 - This should fix #53 due to reverting the changes and allowing tags to be created as `Record` objects as prior to #41.

I believe these changes should properly resolve all related bugs.


tests/integration/test_ipam.py
```diff
@@ -8,7 +8,15 @@ def test_ip_address_nat_inside_outside_correct_objects(nb_client):
...
     ip_inside_refresh = nb_client.ipam.ip_addresses.get(address="192.0.2.1/32")
 
-    assert isinstance(ip_outside.nat_inside, IpAddresses)
-    assert isinstance(ip_inside_refresh.nat_outside, IpAddresses)
     assert str(ip_outside.nat_inside) == "192.0.2.1/32"
     assert str(ip_inside_refresh.nat_outside) == "192.0.2.2/32"
+
+
+def test_prefixes_successfully_stringify_tags(nb_client):
+    """Validate nat_inside and nat_outside both return IpAddress Record objects."""
```
glennmatthews, line 16: Wrong docstring. :-)

pynautobot/models/dcim.py
```diff
@@ -81,9 +80,6 @@ class Devices(Record):
 
     has_details = True
     device_type = DeviceTypes
-    primary_ip = IpAddresses
-    primary_ip4 = IpAddresses
-    primary_ip6 = IpAddresses
```
glennmatthews, line 86: Just to make sure I understand correctly - this is because the only thing that distinguished `IpAddresses` from the parent `Record` class was the custom `__str__` method, which is no longer needed, and therefore there's no longer any need to even have an `IpAddresses` class?

glennmatthews (approved):
Looks appropriate to within the limits of my understanding of this library. :-)

FragmentedPacket:
@thatfatguypat Do you mind testing this PR and make sure it resolves your issue?

pynautobot/models/dcim.py
jmcgill298: what does the `display` attribute represent, and what scenarios do we expect it not to exist?

glennmatthews: > what does the `display` attribute represent, and what scenarios do we expect it not to exist?

In theory all model serializers in Nautobot should provide a `display` attribute that provides a relevant string representation of the object. In practice, it looks like some existing serializers have a gap here - see https://github.com/nautobot/nautobot/pull/1783#discussion_r875896753 for a list

glennmatthews: Sure, see https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/response.py#L212 - could re-add the fallback logic that was removed in #42 and have it try `name` and `label` if there's no `display`...?

jmcgill298: ya, I wasn't really sure if that was still appropriate, but if that makes sense, then I'm good with falling back to what was previously there

FragmentedPacket: I added it back within the latest commit so it'll check for **display** -> **name** -> **label** and then default to empty string.

thatfatguypat:
@FragmentedPacket just gave this a test with my org's automation and it is working the same as prior to #41.  Thanks for putting this together!

PR #58, opened 2022-05-21
WIP: Initial testing strategy for testing custom record __str__ methods

FragmentedPacket:
I'll take the credit of finding (introducing) these bugs :P 

But in reality, we just need to extend test coverage as much as we can within this library so it'll help us understand the underlying code for this library and help reduce regression bugs with changes implemented since the code has changed very little since the fork.

These tests that I'm writing are to mainly make sure that all endpoint records (custom and not) display the correct string and any custom attributes for child objects on records are the correct instances. This is an initial pattern for these tests that I want to get feedback on and make sure it makes sense before moving forward to writing the remaining tests for all endpoints.

### DIFF
--- a/tests/fixtures/circuits/circuit_terminations.json
+++ b/tests/fixtures/circuits/circuit_terminations.json
@@ -1,46 +1,3007 @@
 {
-    "count": 1,
+    "count": 224,
     "next": null,
     "previous": null,
     "results": [
         {
-            "id": 1,
+            "id": "ad666b74-10cf-4be2-a248-0393200a26f3",
+            "url": "https://demo.nautobot.com/api/circuits/circuit-terminations/ad666b74-10cf-4be2-a248-0393200a26f3/",
             "circuit": {
-                "id": 1,
-                "url": "http://localhost:8000/api/circuits/circuits/1/",
-                "cid": "123456"
+                "id": "5b1d5b48-67b7-4eb4-bc72-7623aa1da9b8",
+                "url": "https://demo.nautobot.com/api/circuits/circuits/5b1d5b48-67b7-4eb4-bc72-7623aa1da9b8/",
+                "cid": "ntt-104265404093023273",
+                "display": "ntt-104265404093023273"
             },
             "term_side": "A",
             "site": {
-                "id": 1,
-                "url": "http://localhost:8000/api/dcim/sites/1/",
-                "name": "TEST1",
-                "slug": "test1"
+                "id": "73853b84-2970-4da9-ba6d-2e8224f293fc",
+                "url": "https://demo.nautobot.com/api/dcim/sites/73853b84-2970-4da9-ba6d-2e8224f293fc/",
+                "name": "SIN01",
+                "slug": "sin01",
+                "display": "SIN01"
             },
-            "interface": {
-                "id": 4,
-                "device": {
-                    "id": 1,
-                    "url": "http://localhost:8000/api/dcim/devices/1/",
-                    "name": "test1-edge1",
-                    "display_name": "test1-edge1"
-                },
-                "name": "xe-0/0/0",
-                "form_factor": {
-                    "value": 1200,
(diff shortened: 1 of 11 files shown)


FragmentedPacket:
@glennmatthews I would love your feedback on this as well!

FragmentedPacket:
I think I could potentially fold this into the generic tests actually. I will look into that as well.

tests/fixtures/circuits/circuit_terminations.json
```diff
@@ -1,46 +1,3007 @@
 {
```
glennmatthews, line 1: Do we really need a 3000+ line fixture file? I'm in favor of updating *all* of the fixtures with up-to-date Nautobot data instead of the old quasi-NetBox data, but this seems like a lot more records here than our tests actually need.

tests/unit/test_endpoint.py
```diff
@@ -51,3 +59,57 @@ def test_choices(self):
...
+    app = getattr(pynautobot_api, app)
+    ep = getattr(app, endpoint)
+
+    # Load in the fixture data to be mocked
+    with open(f"tests/fixtures/{fixture}") as data:
+        response = json.loads(data.read())
+
+    # Gather all endpoint records with mocked data
+    with requests_mock.Mocker() as mock_resp:
+        mock_resp.register_uri("GET", f"{ep.url}/", json=response)
```
glennmatthews, line 83: Is this a bug? Shouldn't `ep.url` already include a trailing slash?

tests/unit/test_endpoint.py
```diff
@@ -51,3 +59,57 @@ def test_choices(self):
...
+
+        # Validate correct sub records existing that are defined on custom pynautobot models
+        sub_records = {attr for attr in inspect.getmembers(record, inspect.isclass) if not attr[0] == "__class__"}
+        if sub_records:
+            assert all(
+                [
+                    sub_rec
+                    for sub_rec in sub_records
+                    if not isinstance(getattr(rec, sub_rec[0]), list)
+                    and isinstance(getattr(rec, sub_rec[0]), sub_rec[1])
```
glennmatthews, line 113: Comments would be good here to explain the logic

tests/unit/test_endpoint.py
```diff
@@ -51,3 +59,57 @@ def test_choices(self):
...
+    with requests_mock.Mocker() as mock_resp:
+        mock_resp.register_uri("GET", f"{ep.url}/", json=response)
+        all_records = ep.all()
+
+    # Iterate over records and assert string methods return correctly
+    for rec in all_records:
+        # Validate the specified field is available on the object
+        # It will attempt to make a requests call, but we catch the exception and fail
+        try:
+            parent, *children = field.split(".")
```
glennmatthews, line 91: Comments here with examples to explain the logic would be helpful.

tests/unit/test_endpoint.py
```diff
@@ -51,3 +59,57 @@ def test_choices(self):
...
+            field_value = getattr(rec, parent)
+
+            # We need to iterate to the end of children to get the field_value for the string assertion
+            if field_value and children:
+                while children:
+                    child_field = children.pop(0)
+                    field_value = getattr(field_value, child_field)
```
glennmatthews, line 98: Perhaps?

```suggestion
            if field_value:
                for child_field in children:
                    field_value = getattr(field_value, child_field)
```

tests/fixtures/circuits/circuit_terminations.json
FragmentedPacket: No we don't. This was just done quickly and copied from API docs call on demo.nautobot.com. I will go through records for each endpoint and make sure we cover the different types fields/data, but that should be handful at most.

FragmentedPacket:
I will prob have to have someone else take this over! Not sure if i'll have time in the near future to complete this.

PR #75, opened 2022-09-01
Revert a bit more changes and override the Record.__str__ 

FragmentedPacket:
The child objects for prefixes (available-prefixes/ips) do not return whole objects and need to account for that.

### DIFF
--- a/pynautobot/models/ipam.py
+++ b/pynautobot/models/ipam.py
@@ -19,7 +19,17 @@ from pynautobot.core.response import Record
 from pynautobot.core.endpoint import DetailEndpoint
 
 
+class IpAddresses(Record):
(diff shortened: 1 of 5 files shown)


pynautobot/models/ipam.py
```diff
@@ -19,7 +19,17 @@
 from pynautobot.core.endpoint import DetailEndpoint
 
 
+class IpAddresses(Record):
+    def __str__(self):
+        parent_record_string = super().__str__()
+        return parent_record_string or str(self.address)
```
jvanderaa, line 25: I'm just not sure why we are returning the parent record here.

jvanderaa (approved):

Issue #83, opened 2022-10-12
API URL Endpoint doesn't handle "/"


mattmiller87:
## Re-producable using:
Local Environment
- python version: 3.10.5
- pynautobot version: 1.2.1

Nautobot Environment
- nautobot version: 1.4.6b1
- data validataion plugin version: 1.0.0

## Example
In Nautobot, Create a Regex Rule:
- Name: Test
- Enabled: False
- Content type: dcim.site
- Field: name
- Regular expression: ^[A-Z]

In Pynautobot, update the regex rule to `Enabled: True`
```
import os
import logging
logging.basicConfig()
logging.getLogger().setLevel(logging.DEBUG)
import pynautobot
import requests
url = "http://localhost:8080"
token = "ba21aea2b4f83376c3cfc8b9237660e72040dd83"
requests.packages.urllib3.disable_warnings(requests.packages.urllib3.exceptions.InsecureRequestWarning)
nautobot = pynautobot.api(url=url, token=token)
endpoint = getattr(nautobot.plugins, "data-validation-engine")
endpoint = getattr(endpoint, "rules/regex")
nautobot.http_session.verify = False
endpoint.url
```
at this point we see endpoint.url = 'http://localhost:8080/api/plugins/data-validation-engine/rules/regex'
```
rgs = endpoint.all()
rgs[0].endpoint.url
```
at this point we see rgs[0].endpoint.url = 'http://localhost:8080/api/plugins/data-validation-engine/rules'
```
rgs[0].serialize()
rgs[0].enabled = True
rgs[0].save()
```

error: 
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mattmiller/Documents/playground/env/lib/python3.10/site-packages/pynautobot/core/endpoint.py", line 109, in all
    return response_loader(req.get(), self.return_obj, self)
  File "/Users/mattmiller/Documents/playground/env/lib/python3.10/site-packages/pynautobot/core/query.py", line 326, in get
    return req_all()
  File "/Users/mattmiller/Documents/playground/env/lib/python3.10/site-packages/pynautobot/core/query.py", line 284, in req_all
    req = self._make_call(add_params=add_params)
  File "/Users/mattmiller/Documents/playground/env/lib/python3.10/site-packages/pynautobot/core/query.py", line 257, in _make_call
    raise RequestError(req)
[...]

PR #84, opened 2022-10-19
Fixes plugin nested endpoints

joewesch:
This fixes when a plugin has nested endpoints. For example:
```
https://example.com/api/plugins/app_name/endpoint_name/nested_endpoint_name/<uuid>/
```

Resolves: #83


jvanderaa (approved):
LGTM based on:

* Test case written that has the new URL endpoint
* Existing tests pass that test without a nested endpoint

jvanderaa:
@joewesch I think we are good to merge and then make a patch release.

PR #85, opened 2022-10-20
Add a `retries` option to allow more resilient operation

chadell:
This PR proposes to add a **retry** logic to the SDK, so it will be more "persistent" in front of temporal issues on the server.
This should help to mitigate issue [#16](https://github.com/nautobot/nautobot-ansible/issues/169)

If the idea is welcomed, I would complete the PR adding proper testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 ![pynautobot](docs/nautobot_logo.svg "Nautobot logo")
 
 # Pynautobot
+
 Python API client library for [Nautobot](https://github.com/nautobot/nautobot).
 
 > Pynautobot was initially developed as a fork of [pynetbox](https://github.com/digitalocean/pynetbox/).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -121,6 +122,21 @@ tags.create(name="Tag", slug="tag", api_version="1.2",)
 tags.list(api_version="1.3",)
 ```
 
+### Retry logic
+
+By default, the client will not retry any operation. This behavior can be adjusted via the `retries` optional parameters. This will only affect for HTTP codes: 429, 501, 502, 503 and 504.
```
chadell, line 127:
```suggestion
By default, the client will not retry any operation. This behavior can be adjusted via the `retries` optional parameters. This will only affect for HTTP codes: 429, 500, 502, 503 and 504.
```

chadell:
@jvanderaa @jmcgill298 anything else I should do in this PR?

jvanderaa (approved):

Issue #63, opened 2022-06-24
Update CI Matrix


jvanderaa:
Update the CI to be a better matrix of the Nautobot and Python versions

https://github.com/nautobot/pynautobot/pull/59

techytushar:
Hey @jvanderaa ,
I can pick this up

Issue #79, opened 2022-10-01
invoke start is failing


jifox:
I used this configuration with an unmodified `develompent/dev.env`

```
export NAUTOBOT_VER=1.4.4
export PYTHON_VER=3.7
```

When executing `invoke start` the following traceback is displayed.

```
 *  Executing task in folder nautobot-plugindev: docker logs --tail 1000 -f 6c66bd41055f812cce1490f6d181d5f9203d5aaadbc891094775a71332a0a021 

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/nautobot/extras/plugins/utils.py", line 64, in load_plugin
    plugin = importlib.import_module(plugin_name)
  File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'example_plugin'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/nautobot/core/runner/runner.py", line 120, in settings_callback
    "settings": settings,
  File "/usr/local/lib/python3.7/site-packages/nautobot/core/cli.py", line 176, in _configure_settings
    load_plugins(settings)
  File "/usr/local/lib/python3.7/site-packages/nautobot/extras/plugins/utils.py", line 54, in load_plugins
    load_plugin(plugin_name, settings)
  File "/usr/local/lib/python3.7/site-packages/nautobot/extras/plugins/utils.py", line 70, in load_plugin
    ) from err
nautobot.extras.plugins.exceptions.PluginNotFound: Unable to import plugin example_plugin: Module not found. Check that the plugin module has been installed within the correct Python environment.
⏳ Waiting on DB... (0s / 30s)
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/nautobot/extras/plugins/utils.py", line 64, in load_plugin
    plugin = importlib.import_module(plugin_name)
  File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
[...]

PR #87, opened 2022-10-20
ci: update github ci action

techytushar:
Signed-off-by: Tushar Mittal <chiragmittal.mittal@gmail.com>

Fixes: #63

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,13 +10,11 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        include:
-          - python-version: "3.7"
-            nautobot-version: "1.2"
-          - python-version: "3.8"
-            nautobot-version: "1.3"
-          - python-version: "3.9"
-            nautobot-version: "1.3"
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        nautobot-version: ["1.2", "1.3"]
+        exclude:
+          - python-version: 3.10
+            nautobot-version: 1.2
     env:
       PYTHON_VER: "${{ matrix.python-version }}"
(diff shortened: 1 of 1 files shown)


joewesch:
Hey @techytushar, thanks for the submission! Would you mind adding Python version 3.10 and Nautobot version 1.4 to the matrix as well?

techytushar:
@joewesch Done ✅

.github/workflows/ci.yml
```diff
@@ -10,13 +10,8 @@ jobs:
...
       matrix:
-        include:
-          - python-version: "3.7"
-            nautobot-version: "1.2"
-          - python-version: "3.8"
-            nautobot-version: "1.3"
-          - python-version: "3.9"
-            nautobot-version: "1.3"
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
+        nautobot-version: ["1.2", "1.3", "1.4"]
```
joewesch, line 14: Looks like we need to exclude a combination that doesn't exist.

```suggestion
        nautobot-version: ["1.2", "1.3", "1.4"]
        exclude:
          - python-version: 1.2
            nautobot-version: 3.10
```

jvanderaa:
@joewesch any thoughts on the failure within the tests? Do we just need to re-run it?

joewesch:
The `nautobot-dev:1.2-py3.10` container tag does not exist, so we need to exclude it from the matrix. I added a suggestion that would do that.

joewesch:
Now it's getting hit with the same bug as #79. @techytushar, sorry but can you remove Nautobot 1.4 from the tests for now. We will need to override the nautobot_config.py file if we want that version to run.

joewesch (approved):

PR #90, opened 2022-11-02
Docker-compose fixes

glennmatthews:
- Since we're overriding the default in the base Docker image to run on a different port than the default, `nautobot-server runserver` needs the `--insecure` option to serve static files successfully
- RQ worker shouldn't be needed these days.


jvanderaa:
@glennmatthews I think we need the celery worker swap in here.

glennmatthews:
Possibly so - are we doing anything with pynautobot at present that needs the worker though?

jvanderaa:
Nothing, just meant as a test unit.

jvanderaa (approved):

PR #98, opened 2022-12-24
Handle app names including dash -

johanek:
Currently access app names including a - character results in an error, as underscores are not converted to dashes i.e.:

```
In [2]: nb.plugins.golden_config.config_compliance.all()
...
RequestError: The requested url: https://nautobot-staging.internal/api/plugins/golden_config/config-compliance/?limit=0 could not be found.
```

### DIFF
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -141,7 +141,7 @@ class PluginsApp(object):
         self.api = api
 
     def __getattr__(self, name):
-        return App(self.api, "plugins/{}".format(name))
+        return App(self.api, f"plugins/{name.replace('_', '-')}")
 
     def installed_plugins(self):
         """Returns raw response with installed plugins
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -50,14 +50,13 @@ class Endpoint(object):
 
     def __init__(self, api, app, name, model=None):
         self.return_obj = self._lookup_ret_obj(name, model)
-        self.app_name = app.name.replace("_", "-")
         self.name = name.replace("_", "-")
         self.api = api
         self.base_url = api.base_url
         self.token = api.token
         self.url = "{base_url}/{app}/{endpoint}".format(
             base_url=self.base_url,
-            app=self.app_name,
+            app=app.name,
             endpoint=self.name,
         )
         self._choices = None
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -50,13 +50,14 @@ class Endpoint(object):
 
     def __init__(self, api, app, name, model=None):
         self.return_obj = self._lookup_ret_obj(name, model)
+        self.app_name = app.name.replace("_", "-")
         self.name = name.replace("_", "-")
         self.api = api
         self.base_url = api.base_url
         self.token = api.token
         self.url = "{base_url}/{app}/{endpoint}".format(
             base_url=self.base_url,
-            app=app.name,
+            app=self.app_name,
             endpoint=self.name,
         )
         self._choices = None


jvanderaa:
Will check the code. If I remember back a bit, something like this caused some issues in the code and was changed to what it is now.

There is also this discussion that outlines the current methodology (if this PR is not an improvement on it): https://github.com/nautobot/pynautobot/discussions/62

nautics889 (approved):
Verified the patch, works correctly.
___
nautobot    1.5.14
nautobot-golden-config    1.3.1
nautobot-plugin-nornir    1.0.0
___
Before (current implementation from `HEAD` of `develop` branch):
```
>>> import os
>>> from pynautobot import api
>>> nb_api = api("http://localhost:8888", token=os.environ.get("NAUTOBOT_API_TOKEN"))
>>> nb_api.plugins.golden_config.config_compliance.all()
  File "C:\Program Files\JetBrains\PyCharm 2022.3.2\plugins\python\helpers\pydev\pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
  File "C:\Users\cyber\pynautobot\pynautobot\core\endpoint.py", line 109, in all
    return response_loader(req.get(), self.return_obj, self)
  File "C:\Users\cyber\pynautobot\pynautobot\core\query.py", line 345, in get
    return req_all()
  File "C:\Users\cyber\pynautobot\pynautobot\core\query.py", line 303, in req_all
    req = self._make_call(add_params=add_params)
  File "C:\Users\cyber\pynautobot\pynautobot\core\query.py", line 276, in _make_call
    raise RequestError(req)
pynautobot.core.query.RequestError: The requested url: http://localhost:8888/api/plugins/golden_config/config-compliance/ could not be found.
```
___

After fetching the patch:
```
>>> import os
>>> nb_api = api("http://localhost:8888", token=os.environ.get("NAUTOBOT_API_TOKEN"))
  File "C:\Program Files\JetBrains\PyCharm 2022.3.2\plugins\python\helpers\pydev\pydevconsole.py", line 364, in runcode
    coro = func()
  File "<input>", line 1, in <module>
NameError: name 'api' is not defined
>>> from pynautobot import api
>>> nb_api = api("http://localhost:8888", token=os.environ.get("NAUTOBOT_API_TOKEN"))
>>> nb_api.plugins.golden_config.config_compliance.all()  # empty list is OK, i hadn't created any
[]
```
___
As for the code, i have no question.

jmcgill298:
what if the app name uses underscores instead of dashes?

nautics889:
@jmcgill298 

> what if the app name uses underscores instead of dashes?

You're right. This PR works only for specific case, because the plugin Golden Config [sets them `base_url` with dashes](https://github.com/nautobot/nautobot-plugin-golden-config/blob/e40fbfff3279c42b11efee7b80d4492330ae25d9/nautobot_golden_config/__init__.py#L23).

Meanwhile in endpoint.py module the URL for request [is formed with app's `name`, not app's `base_url`](https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/endpoint.py#L59). I guess that is the point, that's the reason why the original request failed with 404. I mean, it'd work until the `name` of plugin app and it's `base_url` differed. And this is exactly the case @johanek encountered on.

nautics889:
I think i've discovered the very root of the issue.
___

TL;DR We should get `base_url` of a plugin app and use it to form URL rather than use app's package name.

___

When we're operating with any plugin application (like `nb.plugins.golden_config` e. g.),[ the name of attribute (i. e. the name of the plugin) is being used to form the URL](https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/app.py#L146) (just like i've mentioned above). 

_pynautobot/core/app.py:_
```
...
class PluginsApp(object):
    ...
    def __getattr__(self, name):
        return App(self.api, "plugins/{}".format(name))
    ...
...
```

And that's why if you made something like this, using dashes instead of underscores:
```
>>> getattr(nb.plugins, 'golden-config').config_compliance.all()  # it doesn't raise an error
[]
```
this wouldn't return 404, this would be a valid request.

I presume this is incorrect logic. We should use [actual URL](https://github.com/nautobot/nautobot-plugin-golden-config/blob/e40fbfff3279c42b11efee7b80d4492330ae25d9/nautobot_golden_config/__init__.py#L23) of the plugin application to form URL for API-request. Meanwhile the plugin app should be accessible by it's [package name](https://github.com/nautobot/nautobot-plugin-golden-config/blob/e40fbfff3279c42b11efee7b80d4492330ae25d9/nautobot_golden_config/__init__.py#L17). 

___

To make it clear. The behaviour we should reach, example:
app's base url: `golden-config` (see [here](https://github.com/nautobot/nautobot-plugin-golden-config/blob/e40fbfff3279c42b11efee7b80d4492330ae25d9/nautobot_golden_config/__init__.py#L23))
app's package name: `nautobot_golden_config` (see [here](https://github.com/nautobot/nautobot-plugin-golden-config/blob/e40fbfff3279c42b11efee7b80d4492330ae25d9/nautobot_golden_config/__init__.py#L17))
The code which supposed to access plugin app:
```
>>> nb.plugins.nautobot_golden_config.whatever_endpoint_you_want.all()
# and it creates a request to https://<nb_host>/api/plugins/golden-config/config-compliance
```
___

I've investigated `PluginsApp` and it's [`installed_plugins()` method](https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/app.py#L148) to discover how we could get `base_url` of the plugin app. However i encountered that **Nautobot does not return plugin's `base_url` at all**. Here what i've got accessing `/api/plugins/installed-plugins/`:
```
GET /api/plugins/installed-plugins/
HTTP 200 OK
API-Version: 1.2
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

[
    {
        "name": "Nautobot Plugin for Nornir",
        "package": "nautobot_plugin_nornir",
        "author": "Network to Code, LLC",
        "author_email": "",
        "description": "A plugin/library for using Nornir within Nautobot.",
        "verison": "1.0.0",
        "version": "1.0.0"
    },
    {
        "name": "Golden Configuration",
        "package": "nautobot_golden_config",
        "author": "Network to Code, LLC",
        "author_email": "[opensource@networktocode.com](mailto:opensource@networktocode.com)",
        "description": "A plugin for managing Golden Configurations.",
        "verison": "1.3.1",
        "version": "1.3.1"
    }
]
```
I don't see plugin's `base_url` above. 

___

Then i've dove into Nautobot's sources to make clear what `/api/plugins/installed-plugins/` can show. [Here is `_get_plugin_data()` staticmethod](https://github.com/nautobot/nautobot/blob/388e29f35e32440fb3e52dc3c7c810c0f4e634a0/nautobot/extras/plugins/views.py#L101) of the class `InstalledPluginsAPIView`. Here is what it looks like:
```
    @staticmethod
    def _get_plugin_data(plugin_app_config):
        return {
            "name": plugin_app_config.verbose_name,
            "package": plugin_app_config.name,
            "author": plugin_app_config.author,
            "author_email": plugin_app_config.author_email,
            "description": plugin_app_config.description,
            # 2.0 TODO: Remove verison key/value when bumping to major revision
            "verison": plugin_app_config.version,
            "version": plugin_app_config.version,
        }
```
There is no key `base_url` in the result dict as you can see, however it can be added: `"base_url": plugin_app_config.base_url`. Once i added one, the plugin's `base_url` appeared in the `/api/plugins/installed-plugins/`: 
```
GET /api/plugins/installed-plugins/
HTTP 200 OK
API-Version: 1.2
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

[
    {
        "name": "Nautobot Plugin for Nornir",
        "base_url": "plugin-nornir",
        "package": "nautobot_plugin_nornir",
        "author": "Network to Code, LLC",
        "author_email": "",
        "description": "A plugin/library for using Nornir within Nautobot.",
        "verison": "1.0.0",
        "version": "1.0.0"
    },
    {
        "name": "Golden Configuration",
        "base_url": "golden-config",
        "package": "nautobot_golden_config",
        "author": "Network to Code, LLC",
        "author_email": "[opensource@networktocode.com](mailto:opensource@networktocode.com)",
        "description": "A plugin for managing Golden Configurations.",
        "verison": "1.3.1",
        "version": "1.3.1"
    }
]
```
And then we are able to know the actual URL of a plugin.

___

In conclusion. My proposal:

1. Open an issue in https://github.com/nautobot/nautobot/, suggest to update [`_get_plugin_data()`](https://github.com/nautobot/nautobot/blob/388e29f35e32440fb3e52dc3c7c810c0f4e634a0/nautobot/extras/plugins/views.py#L101) with adding `base_url` like this:
```
class InstalledPluginsAPIView(NautobotAPIVersionMixin, APIView):
    ...

    @staticmethod
    def _get_plugin_data(plugin_app_config):
        return {
            "name": plugin_app_config.verbose_name,
            "base_url": plugin_app_config.base_url,
            "package": plugin_app_config.name,
            ...
        }

    ...
```

2. Update [`__getattr__()`](https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/app.py#L145) method of `PluginsApp` class, make something like this:
```
    def __getattr__(self, name):
        try:
            request_for_plugin = Request(
                base="{}/plugins/{}".format(
                    self.api.base_url,
                    name
                ),
                token=self.api.token,
            )
            # make sure the plugin can be accessed by it's package name
            request_for_plugin._make_call()
        except RequestError as error:
            # if not, iterate over installed plugin to try to find `base_url` of the plugin app
            if error.req.status_code == 404:
                for plugin in self.installed_plugins():
                    if plugin.get('package') == name and 'base_url' in plugin:
                        return App(self.api, "plugins/{}".format(plugin['base_url']))
                else:
                    raise error

        return App(self.api, "plugins/{}".format(name))
```
I know, the code above may seem like a weird solution, since it causes to make an additional GET-request to plugin each time we access `nb.plugins.some_plugin_of_mine`, but that's the first came to mind, IDK how else we can check the plugin isn't accessible by it's package name and has different URL.

The code above works well:
```
>>> nb = api("http://localhost:8888", token=os.environ.get("NAUTOBOT_API_TOKEN"))
>>> print(nb.plugins.nautobot_golden_config.config_compliance.all())
[]
```

___


@jmcgill298, @jvanderaa what you think?

nautics889 (changes requested):
@johanek 
Simple replacing underscores with dashes is an incorrect solution.
I'd suggest you to use something like this in your client code:
```
>>> getattr(nb.plugins, 'golden-config').config_compliance.all()
[]
```
or provide some changes according to points i've described.

jvanderaa:
@pszulczewski is also taking a look at this some.

pszulczewski:
Great input @johanek and @nautics889!

As mentioned before the solution proposed by @johanek may affect other app names not only plugins.

Solutions proposed by @nautics889 introduce deviation to the api path documented in API Docs [Swagger](https://demo.nautobot.com/api/docs/#/plugins) as you use plugin names and not the exact plugin `base_url` on `pynautobot` objects. It can also be a long shot to update nautobot with extra methods.

I would be leaning towards replacing `_` to `-` but I would make it only for plugins, so `PluginsApp` `__getattr__`:
https://github.com/nautobot/pynautobot/blob/4c3bad3a65dd0d746004dd638e8a33207ee46271/pynautobot/core/app.py#L146
becomes:
`        return App(self.api, f"plugins/{name.replace('_', '-')}")`

Then we only affect plugin names and we fix something what doesn't really work now.
And that works with the base-url as documented in the API Docs.
```
nb_demo.plugins.golden_config.config_compliance.all()
[<pynautobot.core.response.Record ('jcy-bb-01.infra.ntc.com -> Cisco IOS - svc -> True') ...

nb_demo.plugins.firewall.address_object.all()
[<pynautobot.core.response.Record ('nat-destination') ...

nb_demo.plugins.nautobot_device_lifecycle_mgmt.cve.all()
[<pynautobot.core.response.Record ('CVE-2020-3508') at 0x10294f7f0>]
```

@jvanderaa I am not aware of previous issues you mentioned, what do think about this approach, to replace underscores to dashes only or plugin paths, which are base-urls?

nautics889:
@pszulczewski thank you!

Replacing `_` to `-` seems like good solution for now at least.

FYI: i've created [topic in Nautobot's discussions](https://github.com/nautobot/nautobot/discussions/3622) related to this problem, so we can return to this further.

pszulczewski:
@johanek fancy updating this PR with my suggestion above?

johanek:
@pszulczewski Updated per your suggestion. Let me know if you'd like me to clean the git history

pszulczewski (approved):
That's fine, we can squash upon merge.
Thanks for that.

pszulczewski:
There is a dependency issue with urllib3 as I see container pulls 2.0.2, where we have 1.26.13 specified in poetry.lock.
Looking into this issue ...

pszulczewski (approved):
I see there is re-base option available to me.

Issue #106, opened 2023-02-23
Allow for manual pagination


Kircheneer:
Currently, explicit pagination on the  `filter()` and `all()` endpoints is explicitly forbidden ([ref](https://github.com/nautobot/pynautobot/blob/3bb451340c8b77ba6fdc3b67fa1ea23d23e8f19f/pynautobot/core/endpoint.py#L231)). This can lead to problems where the default timing plus page size (50 when I tried it against demo.nautobot.com) is too much load for the system, causing timeouts or similar errors to happen. This paging behavior happens [here](github.com/nautobot/pynautobot/blob/develop/pynautobot/core/query.py#L289).

My suggestion would be to keep this default behavior, but allow the user to optionally control paging (i.e. pass limit + offset) themselves so they can control their request size + timing with values that work on their system.

PR #107, opened 2023-02-23
Draft for manual paging implementation.

Kircheneer:
Ref: https://github.com/nautobot/pynautobot/issues/106

### DIFF
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -15,6 +15,8 @@ limitations under the License.
 
 This file has been modified by NetworktoCode, LLC.
 """
+import urllib.parse
+
 from pynautobot.core.query import Request, RequestError
 from pynautobot.core.response import Record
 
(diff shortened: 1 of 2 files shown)


Kircheneer:
Should look at pushing the entire logic into `Endpoint`, allowing `limit` and `offset` in kwargs instead. Keeps the API for the end user the same but should be cleaner internally. Also removes the special treatment for `limit` and `offset` as much as possible, as in the end they are just normal REST API parameters.

pynautobot/core/endpoint.py
```diff
@@ -232,13 +254,19 @@ def filter(self, *args, api_version=None, **kwargs):
             raise ValueError("A reserved {} kwarg was passed. Please remove it " "try again.".format(RESERVED_KWARGS))
 
         api_version = api_version or self.api.api_version
+
+        manual_pagination = self._check_manual_pagination(limit, offset)
+        url = "{}/".format(self.url)
+        if manual_pagination:
+            url += "?" + urllib.parse.urlencode({"limit": limit, "offset": offset})
```
jvanderaa, line 261: If this sticks around, let's look at passing it to the Requests library with `params`.

pynautobot/core/query.py
```diff
@@ -303,6 +305,8 @@ def req_all():
             req = self._make_call(add_params=add_params)
             if isinstance(req, dict) and req.get("results") is not None:
                 ret = req["results"]
+                if self.manual_pagination:
+                    return ret
```
jvanderaa, line 309: Add some commentary on this, I'm not clear why returning here.

pynautobot/core/endpoint.py
```diff
@@ -80,7 +82,22 @@ def _lookup_ret_obj(self, name, model):
...
+    def _check_manual_pagination(self, limit, offset):
+        """Checks whether manual pagination should be applied.
+
+        :arg limit: Integer value corresponding to the API parameter.
+        :arg offset: Integer value corresponding to the API parameter.
+
+        :Returns: True or false, depending on whether manual pagination is used."""
+        if limit:
+            if not offset:
+                raise ValueError("If limit is passed to the 'all' method, offset needs to be passed as well.")
```
Kircheneer, line 94: This should be `if offset is None`, otherwise it doesn't allow 0.

joewesch:
This PR is really stale and it understandably now has conflicts. Closing to clean up, but feel free to reopen if you want to pick it back up again in the future.

Issue #67, opened 2022-07-13
Update __repr__ to include Class Name


jvanderaa:
Per issue #65...

It would be good to provide a more descriptive `__repr__` to not confuse what the object is that is being viewed.

nautics889:
@jvanderaa please have a look at https://github.com/nautobot/pynautobot/pull/113

I actually thought about adding `id`, but the represetation string would look too as for me.

joewesch:
closed as part of #113

PR #113, opened 2023-04-01
fix: Enhancement of __repr__ method (#67)

nautics889:
Update `Record` class in pynautobot/core/response.py module. Add more detailed info the `__repr__()` method returns.

Example:

_main.py:_
```
def main():
    api_obj = api('http://localhost:8888',
                  token=os.environ.get('NAUTOBOT_API_TOKEN'))

    device = api_obj.dcim.devices.get('***')
    print(repr(device))

    print("=" * 79)

    interfaces_list = api_obj.dcim.interfaces.all()
    pprint(interfaces_list)

```

Output:
```
<pynautobot.models.dcim.Devices object (name='hAP lite router')>
===============================================================================
[<pynautobot.models.dcim.Interfaces object (name='test ethernet1/1 interface (ethernet1/1)')>,
 <pynautobot.models.dcim.Interfaces object (name='test ethernet1/2 interface (ethernet1/2)')>]
```

### DIFF
--- a/pynautobot/core/response.py
+++ b/pynautobot/core/response.py
@@ -206,7 +206,7 @@ class Record(object):
         return getattr(self, "display", None) or getattr(self, "name", None) or getattr(self, "label", None) or ""
 
     def __repr__(self):
-        return "<{}.{} name='{}' at {}>".format(self.__class__.__module__, self.__class__.__name__, self, hex(id(self)))
+        return "<{}.{} ('{}') at {}>".format(self.__class__.__module__, self.__class__.__name__, self, hex(id(self)))
 
     def __getstate__(self):
         return self.__dict__
--- a/pynautobot/core/response.py
+++ b/pynautobot/core/response.py
@@ -84,7 +84,7 @@ class Record(object):
 
     >>> x = nb.dcim.devices.get(1)
     >>> x
-    <pynautobot.models.dcim.Devices name='test1-switch1' at 0x1953d821250>
+    <pynautobot.models.dcim.Devices ('test1-switch1') at 0x1953d821250>
     >>>
 
     Querying a string field.
--- a/pynautobot/core/response.py
+++ b/pynautobot/core/response.py
@@ -80,11 +80,11 @@ class Record(object):
     :examples:
 
     Default representation of the object usually contains object's
-    class and object's name
+    class, object's name and it's location in memory
 
     >>> x = nb.dcim.devices.get(1)
     >>> x
-    <pynautobot.models.dcim.Devices object (name='test1-switch1')>
+    <pynautobot.models.dcim.Devices name='test1-switch1' at 0x1953d821250>
     >>>
 
     Querying a string field.
@@ -206,7 +206,7 @@ class Record(object):
         return getattr(self, "display", None) or getattr(self, "name", None) or getattr(self, "label", None) or ""
 
     def __repr__(self):
-        return "<{}.{} object (name='{}')>".format(self.__class__.__module__, self.__class__.__name__, self)
+        return "<{}.{} name='{}' at {}>".format(self.__class__.__module__, self.__class__.__name__, self, hex(id(self)))
 
     def __getstate__(self):
         return self.__dict__
--- a/pynautobot/core/response.py
+++ b/pynautobot/core/response.py
@@ -79,11 +79,12 @@ class Record(object):
 
     :examples:
 
-    Default representation of the object is usually its name
+    Default representation of the object usually contains object's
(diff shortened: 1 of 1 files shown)


joewesch:
@nautics889 thank you for the PR. Was there any design choices you made on the repr format or are you copying this pattern from somewhere else? I'm not used to seeing the word `object` in a repr, it's normally just `<class_name at memory_location>`.

nautics889:
> @nautics889 thank you for the PR. Were there any design choices you made on the repr format or are you copying this pattern from somewhere else? I'm not used to seeing the word `object` in a repr, it's normally just `<class_name at memory_location>`.

@joewesch i've been considering what `__repr__()` returns in Python3
```
Python 3.8.10 (default, Apr  4 2023, 14:44:36) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> class Bar:
...     pass
... 
>>> b = Bar()
>>> 
>>> b
<__main__.Bar object at 0x7ffabda20100>
>>> repr(b)
'<__main__.Bar object at 0x7ffabda20100>'
```
implying `b` is an _object_ of class `Bar`. 
In Python2.x the word `instance` was used instead:
```
Python 2.7.18 (default, Apr  4 2023, 14:46:24) 
[GCC 9.4.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> class Bar:
...     pass
... 
>>> b = Bar()
>>> b
<__main__.Bar instance at 0x7f9d2b34c190>
```
Although i've missed memory location in the implementation, i can add one if it matters (like `hex(id(self))`)
___
According to [Python docs](https://docs.python.org/3/library/functions.html#repr) `__repr__()` should return whether a string which can fully recreate same object or "a string enclosed in angle brackets that contains the name of the type of the object together with additional information often including the name and address of the object".

I wouldn't say in this case `__repr__()` should return string for re-creating the object. So i'd make the next implemention (without word 'object' like you've suggested):
```
class Record(object):
...
def __repr__(self):
    "<{}.{} at {} (name='{}')>".format(self.__class__.__module__, self.__class__.__name__, hex(id(self)), self)
...

>>> print(repr(obj))  # for example
<pynautobot.models.dcim.Devices at 0x7f9d2b34c190 (name='test1-switch1')>
...
```

I'll update the PR in a few hours...

nautics889:
Done.

The result would look like:
```
<pynautobot.models.dcim.Devices name='hAP lite router' at 0x1953d8bcf40>
```
___
@joewesch could you please have a look at https://github.com/nautobot/pynautobot/pull/113/commits/cc32152c98659eadece8fe77c89f90315f289485?

pynautobot/core/response.py
```diff
@@ -205,7 +206,7 @@ def __str__(self):
         return getattr(self, "display", None) or getattr(self, "name", None) or getattr(self, "label", None) or ""
 
     def __repr__(self):
-        return str(self)
+        return "<{}.{} name='{}' at {}>".format(self.__class__.__module__, self.__class__.__name__, self, hex(id(self)))
```
joewesch, line 209:
```suggestion
        return "<{}.{} ('{}') at {}>".format(self.__class__.__module__, self.__class__.__name__, self, hex(id(self)))
```

I wonder if using `name=` here could be confusing if there isn't a `name` field on the model. Seeing as the `__str__` method returns display, name or label (or an empty string) it may be disingenuous to claim it is always the "name". Simply wrapping it in parentheses seems a happy medium to me unless you want to dynamically change the label to match the attr being returned.

pynautobot/core/response.py
```diff
@@ -79,11 +79,12 @@ class Record(object):
...
     :examples:
 
-    Default representation of the object is usually its name
+    Default representation of the object usually contains object's
+    class, object's name and it's location in memory
 
     >>> x = nb.dcim.devices.get(1)
     >>> x
-    test1-switch1
+    <pynautobot.models.dcim.Devices name='test1-switch1' at 0x1953d821250>
```
joewesch, line 87:
```suggestion
    <pynautobot.models.dcim.Devices ('test1-switch1') at 0x1953d821250>
```

joewesch (changes requested):

pynautobot/core/response.py
nautics889: @joewesch done!
___
Now it looks like this:
```
<pynautobot.models.dcim.Devices ('hAP lite router') at 0x2bae6948f40>
```

joewesch (approved):

nautics889:
@joewesch thanks for review!

PR #115, opened 2023-04-16
Updating methods for custom field choices

nautics889:
Since Nautobot API has two different endpoints for custom fields and custom field choices (_/api/extras/custom-fields/_ and _/api/extras/custom-field-choices/_ respectively), it might seem some ambiguous that we have a method `custom_choices()` which actually sends a request to _/custom-fields/_.

Rename: 
+ `App.custom_choices()` → `App.custom_fields()` (sends a request to _/extras/custom-fields/_)
+ `AppCustomChoicesTestCase` → `AppCustomFieldsTestCase` (tests for sending a request to _/extras/custom-fields/_)

Add: 
+ `App.custom_field_choices()` (sends a request to _/extras/custom-field-choices/_)
+ `AppCustomFieldChoicesTestCase` (tests for sending a request to _/extras/custom-field-choices/_)

___

I also add checking of called URL in the tests. I presume it's a point to be discussed, because without this checking there's a possibility that tests don't display an error when the method being tested sends request to wrong URL. Considering we have mocks all over test module. So as for me, it'd be good to add such assertions in other places of test modules.

### DIFF
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -15,9 +15,13 @@ limitations under the License.
 
 This file has been modified by NetworktoCode, LLC.
 """
+import logging
+
 from pynautobot.core.endpoint import Endpoint, JobsEndpoint
 from pynautobot.core.query import Request
 from pynautobot.models import circuits, dcim, extras, ipam, users, virtualization
+
+logger = logging.getLogger(__name__)
 
 
 class App(object):
@@ -81,6 +85,38 @@ class App(object):
         ).get()
 
         return self._choices
+
+    def custom_choices(self):
+        """Returns custom-fields response from app
+
+        .. note::
+
+            This method is deprecated and will be removed in pynautobot
+            2.0 or newer. Please use `custom_fields()` instead.
+
+        :Returns: Raw response from Nautobot's custom-fields endpoint.
+        :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
+        :Example:
+
+        >>> nb.extras.custom_choices()
+        {'Testfield1': {'Testvalue2': 2, 'Testvalue1': 1},
+         'Testfield2': {'Othervalue2': 4, 'Othervalue1': 3}}
+        """
+        logger.warning(
+            "WARNING: The method 'custom_choices()' will be removed in "
+            "the next major version (2.x) of pynautobot. Please use "
+            "`custom_fields()` instead."
+        )
+
+        custom_fields = Request(
+            base="{}/{}/custom-fields/".format(
+                self.api.base_url,
+                self.name,
+            ),
+            token=self.api.token,
+            http_session=self.api.http_session,
+        ).get()
+        return custom_fields
 
     def custom_fields(self):
         """Returns custom-fields response from app
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -89,17 +89,33 @@ class App(object):
         :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
         :Example:
 
-        >>> custom_fields_list = nb.extras.custom_fields()
-        >>> print(custom_fields_list[0]['label'])
-        Test custom field for rack
-        >>> print(custom_fields_list[0]['content_types'])
-        ['dcim.rack']
+        >>> nb.extras.custom_fields()
+        [
+            {
+                "id": "5b39ba88-e5ab-4be2-89f5-5a016473b53c",
+                "display": "Test custom field",
+                "url": "http://localhost:8000/api/extras/custom-fields/5b39ba88-e5ab-4be2-89f5-5a016473b53c/",
+                "content_types": ["dcim.rack"],
+                "type": {"value": "integer", "label": "Integer"},
+                "label": "Test custom field",
+                "name": "test_custom_field",
+                "slug": "test_custom_field",
+                "description": "",
+                "required": False,
+                "filter_logic": {"value": "loose", "label": "Loose"},
+                "default": None,
+                "weight": 100,
+                "validation_minimum": None,
+                "validation_maximum": None,
+                "validation_regex": "",
+                "created": "2023-04-15",
+                "last_updated": "2023-04-15T17:45:11.839431Z",
+                "notes_url": "http://localhost:8000/api/extras/custom-fields/5b39ba88-e5ab-4be2-89f5-5a016473b53c/notes/",
+            },
+        ]
         """
         custom_fields = Request(
-            base="{}/{}/custom-fields/".format(
-                self.api.base_url,
-                self.name,
-            ),
+            base=f"{self.api.base_url}/{self.name}/custom-fields/",
             token=self.api.token,
             http_session=self.api.http_session,
         ).get()
@@ -112,18 +128,27 @@ class App(object):
         :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
         :Example:
 
-
-        >>> custom_field_choices_list = nb.extras.custom_field_choices()
-        >>> print(custom_field_choices_list[0]['value'])
-        First option
-        >>> print(custom_field_choices_list[0]['field']['name'])
-        test_custom_field
+        >>> nb.extras.custom_field_choices()
+        [
+            {
+                "id": "5b39ba88-e5ab-4be2-89f5-5a016473b53c",
+                "display": "First option",
+                "url": "http://localhost:8000/api/extras/custom-field-choices/5b39ba88-e5ab-4be2-89f5-5a016473b53c/",
+                "field": {
+                    "display": "Test custom field 2",
+                    "id": "5b39ba88-e5ab-4be2-89f5-5a016473b53c",
+                    "url": "http://localhost:8000/api/extras/custom-fields/5b39ba88-e5ab-4be2-89f5-5a016473b53c/",
+                    "name": "test_custom_field_2"
+                },
+                "value": "First option",
+                "weight": 100,
+                "created": "2023-04-15",
+                "last_updated": "2023-04-15T18:11:57.163237Z"
+            },
+        ]
         """
         custom_fields = Request(
-            base="{}/{}/custom-field-choices/".format(
-                self.api.base_url,
-                self.name,
-            ),
+            base=f"{self.api.base_url}/{self.name}/custom-field-choices/",
             token=self.api.token,
             http_session=self.api.http_session,
         ).get()
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -108,15 +108,7 @@ class App(object):
             "`custom_fields()` instead."
         )
 
-        custom_fields = Request(
-            base="{}/{}/custom-fields/".format(
-                self.api.base_url,
-                self.name,
-            ),
-            token=self.api.token,
-            http_session=self.api.http_session,
-        ).get()
-        return custom_fields
+        return self.custom_fields()
 
     def custom_fields(self):
         """Returns custom-fields response from app
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -82,19 +82,45 @@ class App(object):
 
         return self._choices
 
-    def custom_choices(self):
+    def custom_fields(self):
         """Returns custom-fields response from app
 
         :Returns: Raw response from Nautobot's custom-fields endpoint.
         :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
         :Example:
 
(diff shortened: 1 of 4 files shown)


nautics889:
IDK why tests are failing on URL assertion, when i run it locally, those tests are not failing.
Either via `pytest` directy:
```
$ pytest
...
tests/test_app.py::AppCustomFieldsTestCase::test_custom_fields PASSED                                                                                                                                     [  2%]
tests/test_app.py::AppCustomFieldChoicesTestCase::test_custom_field_choices PASSED                                                                                                                        [  2%]
...
```
or via invoke
```
$ export NAUTOBOT_URL="http://localhost:8000"
$ invoke tests
...
tests/test_app.py::AppCustomFieldsTestCase::test_custom_fields PASSED                                                                                                                                     [  2%]
tests/test_app.py::AppCustomFieldChoicesTestCase::test_custom_field_choices PASSED                                                                                                                        [  2%]
...
```
___
Working on it...

nautics889:
Done, tests are fixed.
The problem was that i'd added checking for passed URL like that:
```
expect_url in mock_obj.call_args.args
```
However `.args` and `.kwargs` attributes [are only available from 3.8 version](https://docs.python.org/3.8/library/unittest.mock.html#unittest.mock.Mock.call_args).

That's why tests were failing only on 3.7.

jvanderaa:
@pszulczewski with what you have been seeing, this look good?

pynautobot/core/app.py
```diff
@@ -82,16 +82,18 @@ def choices(self):
...
         :Example:
 
-        >>> nb.extras.custom_choices()
-        {'Testfield1': {'Testvalue2': 2, 'Testvalue1': 1},
-         'Testfield2': {'Othervalue2': 4, 'Othervalue1': 3}}
+        >>> custom_fields_list = nb.extras.custom_fields()
+        >>> print(custom_fields_list[0]['label'])
+        Test custom field for rack
+        >>> print(custom_fields_list[0]['content_types'])
+        ['dcim.rack']
```
pszulczewski, line 96:
```suggestion
        >>> nb.extras.custom_fields()
        [
            {
                "id": "5b39ba88-e5ab-4be2-89f5-5a016473b53c",
                "display": "Test custom field",
                "url": "http://localhost:8000/api/extras/custom-fields/5b39ba88-e5ab-4be2-89f5-5a016473b53c/",
                "content_types": ["dcim.rack"],
                "type": {"value": "integer", "label": "Integer"},
                "label": "Test custom field",
                "name": "test_custom_field",
                "slug": "test_custom_field",
                "description": "",
                "required": False,
                "filter_logic": {"value": "loose", "label": "Loose"},
                "default": None,
                "weight": 100,
                "validation_minimum": None,
                "validation_maximum": None,
                "validation_regex": "",
                "created": "2023-04-15",
                "last_updated": "2023-04-15T17:45:11.839431Z",
                "notes_url": "http://localhost:8000/api/extras/custom-fields/5b39ba88-e5ab-4be2-89f5-5a016473b53c/notes/",
            },
        ]
```
Examples should show what is really returned from the method.
Think about someone who is looking at this for the first time.

pynautobot/core/app.py
```diff
@@ -82,16 +82,18 @@ def choices(self):
...
         """
         custom_fields = Request(
             base="{}/{}/custom-fields/".format(
```
pszulczewski, line 99: Would be good to convert that to f-string. Refactor is welcomed when it simplifies the code.

pynautobot/core/app.py
```diff
@@ -103,6 +105,30 @@ def custom_choices(self):
...
+        :Returns: Raw response from Nautobot's custom-field-choices endpoint.
+        :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
+        :Example:
+
+
+        >>> custom_field_choices_list = nb.extras.custom_field_choices()
+        >>> print(custom_field_choices_list[0]['value'])
+        First option
+        >>> print(custom_field_choices_list[0]['field']['name'])
+        test_custom_field
```
pszulczewski, line 120:
```suggestion
       >>> nb.extras.custom_field_choices()
        [
            {
                "id": "5b39ba88-e5ab-4be2-89f5-5a016473b53c",
                "display": "First option",
                "url": "http://localhost:8000/api/extras/custom-field-choices/5b39ba88-e5ab-4be2-89f5-5a016473b53c/",
                "field": {
                    "display": "Test custom field 2",
                    "id": "5b39ba88-e5ab-4be2-89f5-5a016473b53c",
                    "url": "http://localhost:8000/api/extras/custom-fields/5b39ba88-e5ab-4be2-89f5-5a016473b53c/",
                    "name": "test_custom_field_2"
                },
                "value": "First option",
                "weight": 100,
                "created": "2023-04-15",
                "last_updated": "2023-04-15T18:11:57.163237Z"
            },
        ]
```
Same here, give full example of what this method is returning.

pynautobot/core/app.py
```diff
@@ -103,6 +105,30 @@ def custom_choices(self):
...
+        """
+        custom_fields = Request(
+            base="{}/{}/custom-field-choices/".format(
+                self.api.base_url,
+                self.name,
+            ),
+            token=self.api.token,
+            http_session=self.api.http_session,
+        ).get()
+        return custom_fields
```
pszulczewski, line 130:
```suggestion
        cf_choices = Request(
            base=f"{self.api.base_url}/{self.name}/custom-field-choices/",
            token=self.api.token,
            http_session=self.api.http_session,
        ).get()
        return cf_choices
```

tests/test_app.py
```diff
@@ -3,23 +3,62 @@
...
-        "pynautobot.core.query.Request.get",
-        return_value={"Testfield1": {"TF1_1": 1, "TF1_2": 2}, "Testfield2": {"TF2_1": 3, "TF2_2": 4}},
+        "requests.sessions.Session.get",
+        side_effect=[Response(fixture="extras/custom_fields.json")],
     )
-    def test_custom_choices(self, *_):
+    def test_custom_fields(self, session_get_mock):
         api = pynautobot.api(host, **def_kwargs)
-        choices = api.extras.custom_choices()
+        choices = api.extras.custom_fields()
```
pszulczewski, line 22:
```suggestion
        cfs = api.extras.custom_fields()
```

tests/test_app.py
```diff
@@ -3,23 +3,62 @@
...
 }
 
 
-class AppCustomChoicesTestCase(unittest.TestCase):
+class AppCustomFieldsTestCase(unittest.TestCase):
     @patch(
-        "pynautobot.core.query.Request.get",
-        return_value={"Testfield1": {"TF1_1": 1, "TF1_2": 2}, "Testfield2": {"TF2_1": 3, "TF2_2": 4}},
+        "requests.sessions.Session.get",
+        side_effect=[Response(fixture="extras/custom_fields.json")],
```
pszulczewski, line 18:
```suggestion
        return_value=Response(fixture="extras/custom_fields.json"),
```
`return_value` is good enough, side_effect is for rising exceptions or iterables.

tests/test_app.py
```diff
@@ -3,23 +3,62 @@
...
+
+        session_get_mock.assert_called_once()
+        expect_url = "{}/extras/custom-fields/".format(api.base_url)
+        self.assertGreaterEqual(len(session_get_mock.call_args), 2)
+        url_passed_in_args = expect_url in session_get_mock.call_args[0]
+        url_passed_in_kwargs = expect_url == session_get_mock.call_args[1].get("url")
+        self.assertTrue(url_passed_in_args or url_passed_in_kwargs)
+
+        self.assertIsInstance(choices, list)
         self.assertEqual(len(choices), 2)
```
pszulczewski, line 32:
```suggestion
        self.assertIsInstance(cfs, list)
        self.assertEqual(len(cfs), 2)
```

tests/test_app.py
```diff
@@ -3,23 +3,62 @@
...
-        self.assertEqual(sorted(choices.keys()), ["Testfield1", "Testfield2"])
+        for field in choices:
```
pszulczewski, line 33:
```suggestion
        for field in cfs:
```

tests/test_app.py
```diff
@@ -3,23 +3,62 @@
...
+            self.assertIsInstance(field.get("name"), str)
+            self.assertIsInstance(field.get("content_types"), list)
+            self.assertIsInstance(field.get("slug"), str)
+            self.assertIn("type", field)
+
+
+class AppCustomFieldChoicesTestCase(unittest.TestCase):
+    @patch(
+        "requests.sessions.Session.get",
+        side_effect=[Response(fixture="extras/custom_field_choices.json")],
```
pszulczewski, line 43:
```suggestion
        return_value=Response(fixture="extras/custom_field_choices.json"),
```

pszulczewski (changes requested):
Good work, just some cosmetic changes around examples, mocking and variable names.

nautics889:
@pszulczewski 
Apologies for the delayed response. Thank you for the review!
___
I have uploaded a commit with all the changes you've suggested.
* update examples in docstrings both in `custom_fields()` and `custom_field_choices()` ✓
* use f-strings for the methods instead of `.format()` ✓
* update naming in related test-cases ✓
* use `return_value` instead of `side_effect` for mocks in related test-cases ✓

All CI checks are passed.
I'd appreciate if you could take a look when you have time.
___
p. s. about docstrings: i had inteded to make them as you described at first, but i was afraid that would look too huge for a docstring =)

pynautobot/core/app.py
```diff
@@ -82,22 +82,73 @@ def choices(self):
 
         return self._choices
 
-    def custom_choices(self):
```
joewesch, line 85: Should we have `.custom_choices()` call `self.custom_fields()` so it is backwards compatible?

@jvanderaa @pszulczewski do we have any specific patterns of notifying the end user the method is deprecated?

nautics889: BTW, this is exactly the question that i had gotten, but i thought it depends on versioning. I guess if it's ok to wait new minor feature release (x.**_y_**.z) for this update, it can be as is. Otherwise i'm ready to upload a new commit, containing the original method and both of new, but renamed. Just let me know.

jvanderaa: I don't think we have a specific pattern today @joewesch. I think we should add in something for this.

pszulczewski: I think it would be good to add `.custom_choices()` to call `self.custom_fields()` as @joewesch suggested and we will keep it in 1.x release train and remove in 2.0.
We can also add `logging` and call `logger.warning()` that the method will be deprecated in 2.x.

nautics889: > I think it would be good to add `.custom_choices()` to call `self.custom_fields()` as @joewesch suggested and we will keep it in 1.x release train and remove in 2.0. We can also add `logging` and call `logger.warning()` that the method will be deprecated in 2.x.

Done

pszulczewski (approved):
Brilliant! Thanks @nautics889

pynautobot/core/app.py
```diff
@@ -103,6 +118,78 @@ def custom_choices(self):
         ).get()
         return custom_fields
```
pszulczewski, line 119:
```suggestion
        return self.custom_fields()
```
You need just that after `logger`.

pszulczewski (changes requested):

pynautobot/core/app.py
nautics889: Done ✓

nautics889:
@pszulczewski thank you for the feedback!
___
I've uploaded the last minor changes you've mentioned
>You need just that after logger.

p. s. there is some problem with CI runners (or rather with dependencies being installed for tests as seems to me according to logs). Failed job: https://github.com/nautics889/pynautobot/actions/runs/5660539569/job/15336640442

pszulczewski (approved):

joewesch:
> p. s. there was some problem with CI runners (or rather with dependencies being installed for tests as seems to me according to logs). Failed job: https://github.com/nautics889/pynautobot/actions/runs/5660539569/job/15336640442

Those are the actions in your fork. They are passing in this repo: https://github.com/nautobot/pynautobot/actions/runs/5660540032/job/15336460260

Your branch is 20 commits behind develop, so may be worth pulling in the latest changes....or just disable and ignore your own CI.

joewesch (approved):

nautics889:
> Those are the actions in your fork

Indeed, didn't notice, thank you

___

@jvanderaa i guess we could merge if it looks ok to you

PR #116, opened 2023-04-25
Fix integration tests for 2.0

pszulczewski:
Part of #117

TODO

- [x] Fix `test_ip_address_nat_inside_outside_correct_objects` when NB 2.0 API is updated.


pszulczewski:
#117

jvanderaa (approved):

jvanderaa:
Looking good at this point.

pynautobot/core/api.py
```diff
@@ -75,6 +77,7 @@ def __init__(
         token=None,
         threading=False,
         api_version=None,
+        api_depth=os.getenv("PYNAUTOBOT_API_DEPTH"),
```
jvanderaa, line 80: Should we add a default depth of 1 here? Or did we decide against that?

```suggestion
        api_depth=int(os.getenv("PYNAUTOBOT_API_DEPTH", "1")),
```

I think we will want it as an integer as well? I believe?

pszulczewski: We didn't clearly decide, but I think it's reasonable to set it for `1`.

development/docker-compose.yml
```diff
@@ -33,7 +33,7 @@ services:
     healthcheck:
       interval: "30s"
       timeout: "10s"
-      start_period: "2m"
+      start_period: "5m"
```
glennmatthews, line 36: FYI Nautobot core uses `10m` currently just to be overly safe.

tests/integration/conftest.py
```diff
@@ -145,14 +148,20 @@ def populate_nautobot_object_types(nb_api, devicetype_library_repo_dirpath):
 
     # add device roles
     for device_role_name in DEVICE_ROLE_NAMES:
-        nb_api.dcim.device_roles.create(name=device_role_name, slug=device_role_name.lower().replace(" ", "-"))
+        nb_api.extras.roles.create(
+            name=device_role_name, slug=device_role_name.lower().replace(" ", "-"), content_types=["dcim.device"]
```
glennmatthews, line 152: Shouldn't need a `slug` in 2.0.
```suggestion
            name=device_role_name, content_types=["dcim.device"]
```

tests/integration/test_api_version.py
```diff
@@ -31,7 +31,9 @@ def skipif_version(self, nb_client):
     def tag(self, nb_client):
         """Create a tag."""
         tag_name = "Test Tag"
-        tag = nb_client.extras.tags.create(name=tag_name, slug=tag_name.lower().replace(" ", "_"))
+        tag = nb_client.extras.tags.create(
+            name=tag_name, slug=tag_name.lower().replace(" ", "_"), content_types=["dcim.location"]
```
glennmatthews, line 35:
```suggestion
            name=tag_name, content_types=["dcim.location"]
```

tests/integration/test_dcim.py
```diff
@@ -17,115 +17,137 @@ class TestSimpleServerRackingAndConnecting:
...
-        site = nb_client.dcim.sites.create(name=site_name, slug=site_name.lower().replace(" ", "_"), status="active")
-        assert site
+    def location(self, nb_client):
+        """Verify we can create a location."""
+        location_types = ["dcim.device", "dcim.rack", "dcim.rackgroup"]
+        site_type = nb_client.dcim.location_types.create(name="Site", content_types=location_types, nestable=True)
+        location_name = "MSP"
+        location = nb_client.dcim.locations.create(
+            name=location_name,
+            slug=location_name.lower().replace(" ", "_"),
```
glennmatthews, line 27:
```suggestion
```

tests/integration/test_dcim.py
```diff
@@ -17,115 +17,137 @@ class TestSimpleServerRackingAndConnecting:
...
+            status={"name": "Active"},
+            location_type=site_type.id,
+            depth=3,
```
glennmatthews, line 30: `depth` shouldn't have any effect on create calls, FWIW.

tests/integration/test_dcim.py
```diff
@@ -17,115 +17,137 @@ class TestSimpleServerRackingAndConnecting:
...
 
     @pytest.fixture
-    def data_leafs(self, rack):
+    def data_leafs(self, rack, location):
         """Verify we can create data leaf switch devices."""
         devices = []
         for i in [1, 2]:
             device = rack.api.dcim.devices.create(
                 name=f"access_switch{i}.networktocode.com",
                 device_type={"slug": "dcs-7050tx3-48c8"},
```
glennmatthews, line 61: This needs to change probably to a `model` reference instead of `slug`. Same comment below.

tests/integration/test_ipam.py
```diff
@@ -3,20 +3,26 @@
...
+    assert str(ip_outside_refresh.nat_inside.address) == "192.0.2.1/32"
+    assert str(ip_inside_refresh.nat_outside_list[0].address) == "192.0.2.2/32"
 
 
 def test_prefixes_successfully_stringify_tags(nb_client):
     """Validate prefix will properly stringify the tags attribute and they are Record objects."""
-    tag = nb_client.extras.tags.create(name="production", slug="production")
-    prefix = nb_client.ipam.prefixes.create(prefix="192.0.2.0/24", status="active", tags=[tag.id])
-
+    tag = nb_client.extras.tags.create(name="production", slug="production", content_types=["ipam.prefix"])
```
glennmatthews, line 21:
```suggestion
    tag = nb_client.extras.tags.create(name="production", content_types=["ipam.prefix"])
```

tests/integration/test_dcim.py
pszulczewski: Yes, I was testing that and I forgot to remove.

development/docker-compose.yml
pszulczewski: 5m seems to be fine, locally and for the CI as well. I'll keep an eye on it, in case of issues I will increase it.

tests/integration/conftest.py
pszulczewski: OK, I need to revisit all tests and remove it then.

tests/integration/test_api_version.py
pszulczewski: `pynautobot.core.query.RequestError: The request failed with code 400 Bad Request: {'slug': ['This field is required.']`

glennmatthews: Hmm, may need to wait until 2.0.0b2 is released, this was a fairly recent removal.

jvanderaa (approved):

PR #121, opened 2023-05-11
Pin urllib3 and requests.

pszulczewski:
`requests` doesn't support `urllib3` 2.0 yet see: https://github.com/psf/requests/issues/6432
They don't use poetry for version constraints in `pyproject.toml`, so when our poetry.lock is updated it bumps up `urllib3` to 2.0.2.

Pinning urllib3 in our `pyproject.toml` like in their `setup.cfg`.


pyproject.toml
```diff
@@ -89,4 +91,3 @@ build-backend = "poetry.core.masonry.api"
 [tool.pytest.ini_options]
 testpaths = "tests/"
 addopts = "-vv"
-python_paths = "./"
```
pszulczewski, line 92: That is used for django-pytest plugin and is causing warning in pytest.

joewesch (approved):
Let's make sure to create an issue to track the upstream library and revert this when it is no longer needed.

PR #126, opened 2023-06-13
Add Endpoint.update(id,data) method

### LINKED ISSUE #125: Add method to Endpoint that can directly update an object
I'm working with Nautobot 2.0 and due to the fact that the REST API doesn't resolve related objects has pushed me into the GraphQL world. The code I'm working on needs to review a set of objects and possibly update them. Since I'm already getting all the data back from GraphQL I need, and I know exactly which fields need updating (thanks to diffsync), I'd like to update the objects without adding the unnecessary HTTP GET into the mix.

For example, the way to do it today:
1. Query data from GraphQL
2. Determine updates
3. Load object using `o = nb.dcim.devices.get(device_id)` which results in another GET
4. Make a `o.update(...)` call to call PATCH

This results in an unnecessary GET of every object I want to update. A work-around today is to manually construct a Request to PATCH the object, but this is likely error-prone. I think pynautobot can provide a better API.

Example use of such an API:
```py
# Add a method to class Endpoint.update() that resembles Record.update()
nb.dcim.devices.update(id=..., data={"field1": "value1"})
```

The call would return True if the PATCH was successful and False if unsuccessful.

References:
- [Thread on NTC slack in #nautobot](https://networktocode.slack.com/archives/C01NWPK6WHL/p1686612032204309)

jamesharr:
Adds an `Endpoint.update(id,data)` call to the endpoint API.

Closes #125

### DIFF
--- a/docs/advanced/graphql.rst
+++ b/docs/advanced/graphql.rst
@@ -152,3 +152,12 @@ This example shows accessing data from the previous query.
     >>> # Get the name of the first site
     >>> graphql_response.json["data"]["sites"][0]["name"]
     'den'
+
+
+Saving Changes
+--------------
+
+The Nautobot GraphQL API is currently read-only. To make updates to objects, see:
+
+* :ref:`Updating objects without loading data`
+* :py:meth:`~pynautobot.core.endpoint.Endpoint.update`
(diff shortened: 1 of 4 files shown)


jvanderaa:
I like the plan here. Any chance you would like to take a crack at a test on it? Otherwise I may be able to get that test written in a couple of days here (scheduling, not time to do to the test).

jamesharr:
@jvanderaa 

I took a stab at tests and docs. I'm more confident in the testing than the docs 😆

jamesharr:
Sorry about the weird commit list; I rebased against main, then rebased against develop

joewesch (approved):
Looks good, thanks!

PR #129, opened 2023-06-29
max_workers added 

NobodyIsPerfect78:
max_workers added as a parameter, this allows to set the number of workers when threading=true.

### DIFF
--- a/pynautobot/core/api.py
+++ b/pynautobot/core/api.py
@@ -74,6 +74,7 @@ class Api(object):
         url,
         token=None,
         threading=False,
+        max_workers:int=4,
         api_version=None,
         retries=0,
     ):
@@ -94,6 +95,7 @@ class Api(object):
             self.http_session.mount("http://", _adapter)
             self.http_session.mount("https://", _adapter)
         self.threading = threading
+        self.max_workers = max_workers
         self.api_version = api_version
 
         self.dcim = App(self, "dcim")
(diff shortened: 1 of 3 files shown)


pynautobot/core/api.py
```diff
@@ -74,6 +76,7 @@ def __init__(
         url,
         token=None,
         threading=False,
+        max_workers:int=4,
```
Kircheneer, line 79: I think given that we have type hints still in the docstring and no other type hints in the function signature, maybe we should drop these?

pynautobot/core/query.py
```diff
@@ -131,6 +131,7 @@ def __init__(
         key=None,
         token=None,
         threading=False,
+        max_workers:int=4,
```
Kircheneer, line 134: Same here

Kircheneer (changes requested):
Small nitpick, looks like a rather straight-forward change to me.

pynautobot/core/api.py
NobodyIsPerfect78: fixed

pynautobot/core/query.py
NobodyIsPerfect78: fixed

Kircheneer (approved):
LGTM

jvanderaa (approved):

PR #130, opened 2023-07-11
Add version constraint to divide into two release trains 1.x and 2.x

pszulczewski:


pynautobot/core/api.py
```diff
@@ -109,6 +110,12 @@ def __init__(
         self.users = App(self, "users")
         self.plugins = PluginsApp(self)
         self.graphql = GraphQLQuery(self)
+        self._validate_version()
+
+    def _validate_version(self):
+        """Validate API version if eq or ge than 2.0 raise an error."""
+        if self.version.replace(".", "").isnumeric() and version.parse(self.version) >= version.parse("2.0"):
+            raise ValueError("Nautobot version 2 detected, please upgrade pynautobot to version 2.x")
```
joewesch, line 118: I'm typically not a fan of making API calls as part of an `__init__` method. I don't necessarily have a better solution just yet, but I did also want to point out that I don't think the `self.version` property is cached so this looks to be making 2 calls back to back with `self.version.replace(".", "").isnumeric()` and then `version.parse(self.version)`.

pszulczewski: Yes, this certainly needs an improvement. That was just a quick draft to gather your feedback if we want this.
Many thanks for your feedback. I will assign it to a var inside that method.

joewesch (approved):

Issue #114, opened 2023-04-15
`custom-fields` and `custom-field-choices` are supposed to be different endpoint


nautics889:
In `App` class of pynautobot library there is method for retrieving data from '/api/extras/**custom-fields**/' endpoint. The method has name [`custom_choices()`](https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/app.py#L85). 
And it can make some mess, because Nautobot service actually provide another endpoint "/api/extras/**custom-field-choices**/". So such naming can be somewhat confusing:
_from /api/extras:_
```
Extras API root view

GET /api/extras/


{
    ...
    "custom-field-choices": "http://localhost:8888/api/extras/custom-field-choices/",
    "custom-fields": "http://localhost:8888/api/extras/custom-fields/",
    ...
}
```
And they return completely different data.

Futhermore, there is no method to get data from "/api/extras/**custom-field-choices**/". Perhaps it should be implemented.

PR #144, opened 2023-10-18
Fix (#141)

### LINKED ISSUE #141: custom field endpoint
Since the method custom_fields() was introduced in app.py, you're not able to get the custom_fields endpoint anymore because api.extras.custom_fields pointing to the method custom_fields() of app.py. in Version 1.5 the following code was working:

nautobot = api(url=na_url, token=token)
cf_model = nautobot.extras.custom_fields
cf_model.create(data)

on newer versions, the workaround looks like this:
nautobot = api(url=na_url, token=token)
cf_model = nautobot.extras.__getattr__("custom_fields")
cf_model.create(data)

the same applies to "custom_field_choises"

pszulczewski:
Add `get` to methods to fix `custom_fields` and `custom_field_choices` methods overlap with endpoints.
Fixes #141

### DIFF
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -86,38 +86,14 @@ class App(object):
 
         return self._choices
 
-    def custom_choices(self):
-        """Returns custom-fields response from app
-
-        .. note::
-
-            This method is deprecated and will be removed in pynautobot
-            2.0 or newer. Please use `custom_fields()` instead.
-
-        :Returns: Raw response from Nautobot's custom-fields endpoint.
-        :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
-        :Example:
-
-        >>> nb.extras.custom_choices()
-        {'Testfield1': {'Testvalue2': 2, 'Testvalue1': 1},
-         'Testfield2': {'Othervalue2': 4, 'Othervalue1': 3}}
-        """
-        logger.warning(
-            "WARNING: The method 'custom_choices()' will be removed in "
-            "the next major version (2.x) of pynautobot. Please use "
-            "`custom_fields()` instead."
-        )
-
-        return self.custom_fields()
-
-    def custom_fields(self):
+    def get_custom_fields(self):
         """Returns custom-fields response from app
 
         :Returns: Raw response from Nautobot's custom-fields endpoint.
         :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
         :Example:
 
-        >>> nb.extras.custom_fields()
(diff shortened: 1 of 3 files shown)


pszulczewski:
FYI we had to add `get` to the methods you added as they overlap with endpoint names, if you're using them please append `get`.
//cc @nautics889

nautics889:
@pszulczewski 
___
Apologies for this oversight. Seems like that case has been undertested from my side.
Indeed, if we run something like
```python
import pynautobot
nb = pynautobot.api(
    url="https://demo.nautobot.com/",
    token="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
)

print(repr(nb.extras.custom_fields))
print(repr(nb.extras.__getattr__("custom_fields")))
```
it produces the next output:
```
<bound method App.custom_choices of <pynautobot.core.app.App object at 0x7f233ca46680>>  # this one is what a client actually get, an `App` instance
<pynautobot.core.endpoint.Endpoint object at 0x7f233ca46da0>  # this one is what a client expects to get, an `Enpoint`
```
Meanwhile, with the old implementation of `.custom_fields()`, when it was named [`.custom_choices()`](https://github.com/nautobot/pynautobot/blob/v1.5.0/pynautobot/core/app.py#L85) in v1.5.0 (18f2717e33f4518be2d8bacc0f3068582ca5ecb0) the `Endpoint` instance was available to be gotten by `nb.extras.custom_fields` (because `App` class didn't have this method and `__getattr__` was called then, returning `Endpoint` instance). This point makes me think that these difference in naming ([custom_**choices**](https://github.com/nautobot/pynautobot/blob/v1.5.0/pynautobot/core/app.py#L85) for fetching [custom_**fields**](https://github.com/nautobot/pynautobot/blob/v1.5.0/pynautobot/core/app.py#L97) exactly) has been made intentionally then. 
However, I'm still thinking that the naming approach wasn't fully correct then (in older versions), since ([like i claimed before](https://github.com/nautobot/pynautobot/issues/114#issue-1669526433)) this can cause `/api/extras/custom-fields/` and `/api/extras/custom-field-choices/` to be confused. But i must have discovered that inability of getting custom_fields endpoint.
___
I think the renaming (`custom_fields`→`get_custom_fields` and `custom_field_choices`→`get_custom_field_choices`) is a great solution for those issues: this lets the user to get according `Endpoint` instances and keeps `/api/extras/custom-fields/` & `/api/extras/custom-field-choices/` different.
Thank you a lot for fixing.

PR #145, opened 2023-10-18
Fix #141 LTM

### LINKED ISSUE #141: custom field endpoint

pszulczewski:
Same as #144 but for `ltm-1.6`

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -20,7 +20,7 @@ class AppCustomFieldsTestCase(unittest.TestCase):
     @patch("pynautobot.api.version", "1.999")
     def test_custom_fields(self, session_get_mock):
         api = pynautobot.api(host, **def_kwargs)
-        cfs = api.extras.custom_fields()
+        cfs = api.extras.get_custom_fields()
 
         session_get_mock.assert_called_once()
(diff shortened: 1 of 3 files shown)


tests/test_app.py
```diff
@@ -37,6 +37,25 @@ def test_custom_fields(self, session_get_mock):
             self.assertIsInstance(field.get("slug"), str)
             self.assertIn("type", field)
 
+    @patch(
+        "requests.sessions.Session.get",
+        return_value=Response(fixture="extras/custom_fields.json"),
+    )
+    @patch("pynautobot.api.version", "1.999")
+    def test_custom_choices(self, session_get_mock):
```
pszulczewski, line 45: Added this test to test backwards compatibility.

PR #151, opened 2023-11-12
This is actually 2.0.2

jvanderaa:
Release v2.0.2, v2.0.1 is already out and main/develop branches are out of sync. I'm hopeful that this one will go through without any issues to the `main` branch at which point I will sync `main` to `develop`.


chadell (approved):
LGTM

PR #158, opened 2023-11-26
WiP: Add support for generic auth headers

sklemer1:
Add a `auth_header` attribute that is used verbatim as Authorization header (i.e. without prefixing `Token `). We use this to support oauth2 flows with Bearer tokens (patch for nautobot will come soon).

### DIFF
--- a/pynautobot/core/api.py
+++ b/pynautobot/core/api.py
@@ -53,6 +53,7 @@ class Api(object):
     :param str url: The base URL to the instance of Nautobot you
         wish to connect to.
     :param str token: Your Nautobot token.
+    :param str auth_header: Content of the ``Authorization``-Header, e.g. for Bearer Token.
     :param bool,optional threading: Set to True to use threading in ``.all()``
         and ``.filter()`` requests.
     :param int,optional max_workers: Set the maximum workers for threading in ``.all()``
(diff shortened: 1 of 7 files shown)


joewesch:
I think a better implementation is to just accept `headers` and if it's `None` set the default.
```python
if headers is not None:
    self.headers = headers
else:
    self.headers = {"Authorization": f"Token {self.token}"}
```

sklemer1:
> I'm not seeing requests having an `auth_headers` parameter. Is there something that is undocumented there? https://requests.readthedocs.io/en/latest/api/

Nope, there is no such thing -- but I don't use it like this. The argument `auth_header` is only ever used in pynautobots 'Request' object and as before we hand a complete 'headers' dictionary to the requests request.

> I think a better implementation is to just accept headers and if it's None set the default.

Sure. Would you accept a PR if we write it like this? Would be a tiny little bit more code change as headers are added here and there and the dicts have to be merged then.

joewesch:
> Sure. Would you accept a PR if we write it like this? Would be a tiny little bit more code change as headers are added here and there and the dicts have to be merged then.

Yes. I would suggest using something like `self.headers.update()` for the times we need to merge headers.

joewesch:
@sklemer1 do you plan on resuming this PR at some point or should we close it?

PR #165, opened 2024-02-05
Added bulk update support to the endpoint class

### LINKED ISSUE #164: Modify the Endpoint.update() method to support bulk updates
Many endpoints [allow for bulk update](https://docs.nautobot.com/projects/core/en/stable/user-guide/platform-functionality/rest-api/overview/#updating-multiple-objects) by sending in a list of dicts to the endpoint. We should allow for this feature in pynautobot.

Example usage:

```
import pynautobot
nautobot = pynautobot.api(
  url="http://localhost:8000", token="d6f4e314a5b5fefd164995169f28ae32d987704f",
)
devices = nautobot.dcim.devices
objects = [{"id":"db8770c4-61e5-4999-8372-e7fa576a4f65","name":"test1"},{"id":"e9b5f2e0-4f20-41ad-9179-90a4987f743e","name":"test2"}]
devices.update(objects)
```

vu-smitt13:
This modifies the update method to support bulk updates on the endpoint class along with other changes to support bulk updates. This will resolve #164

### DIFF
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -16,7 +16,6 @@ limitations under the License.
 This file has been modified by NetworktoCode, LLC.
 """
 
-from typing import Dict
 from uuid import UUID
 from pynautobot.core.query import Request, RequestError
 from pynautobot.core.response import Record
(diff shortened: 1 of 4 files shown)


pynautobot/core/endpoint.py
```diff
@@ -311,42 +310,101 @@ def create(self, *args, api_version=None, **kwargs):
...
+        Objects is a list which contain either json/dicts or Record
+        derived objects, which contain the updates to apply.
+        If json/dicts are used, then the id of the object *must* be
+        included
 
-        Accepts the id of the object that needs to be updated as well as
-        a dictionary of k/v pairs used to update an object. The object
-        is directly updated on the server using a PATCH request without
-        fetching object information.
+        :arg str,optional \*args: A list of dicts or a list of Record
```
joewesch, line 323:
```suggestion
        :arg list,optional \*args: A list of dicts or a list of Record
```

pynautobot/core/endpoint.py
```diff
@@ -311,42 +310,101 @@ def create(self, *args, api_version=None, **kwargs):
...
+        [Device1, Device2, Device3]
+        >>> for d in devices:
+        ...     d.name = d.name+'-test'
+        ...
+        >>> nb.dcim.devices.update(devices)
+        >>>
         """
+        objects = args[0] if args else []
+        if not isinstance(objects, list):
+            raise ValueError("objects must be a list[dict()|Record] not " + str(type(objects)))
```
joewesch, line 366: I don't think this would be fully backwards compatible. In the current version, you could technically send the arguments in as positional args instead of keyword args and it would break in this version.

For example, this is technically possible today:
```
nb.dcim.devices.update("0238a4e3-66f2-455a-831f-5f177215de0f", data={"foo": "bar"})
```

Please account for this by checking `len(args)` and splitting them accordingly if the length is 2.

pynautobot/core/response.py
```diff
@@ -361,6 +362,30 @@ def fmt_dict(k, v):
         init = Hashabledict({fmt_dict(k, v) for k, v in self.serialize(init=True).items()})
         return set([i[0] for i in set(current.items()) ^ set(init.items())])
 
+    def updates(self):
```
joewesch, line 365: I know your `Endpoint.update()` tests will technically invoke this method, but we should still have a unit test for this method directly as well.

pynautobot/core/endpoint.py
joewesch: I prefer to keep the status quo with backwards compatibility. If you can incorporate that by using the same method then great, but I'm fine with creating a "bulk" method as well. It does draw the question if we should then rename `.delete()` for consistency (as it hasn't been released yet).

vu-smitt13: Keeping it incorporated with the same method, I came up with this, and it passes those cases.  At some point, the logic inside the method could become too complicated for backwards compatibility.  What do you think about this approach?

```
uuid = ""
data = {}
objects = []
if len(args) == 1 and not kwargs:
    objects = args[0]
    data = []
elif len(args) == 1 and "data" in kwargs:
    uuid = args[0]
    data = kwargs["data"]
elif len(args) >= 2:
    uuid = args[0]
    data = args[1]
elif "data" in kwargs and "id" in kwargs:
    uuid = kwargs["id"]
    data = kwargs["data"]
    
---omitted---
```

joewesch: I think this might look cleaner. Please verify this catches all of the use cases.
```python
    def update(self, *args, **kwargs):
        # Original signature: update(self, id: str, data: Dict[str, any])
        if not args and not kwargs:
            # If we got no args or kwargs, raise a ValueError
            raise ValueError("You must provide either a UUID and data dict or a list of objects to update")
        uuid = kwargs.get("id", "")
        data = kwargs.get("data", {})
        if data and not uuid:
            # If we got data but no uuid, the uuid should be the first arg
            uuid = args[0]
        if len(args) == 2:
            # If we got two args, the first should be the uuid and the second should be the data
            uuid, data = args
        if uuid and data:
            # If we have both, dispatch the original single update method
            < ... >

        # If we got here, we should be dealing with a bulk update
        < ... >
```

vu-smitt13: Combining what you have with some changes I made to reduce code redundancy, it's looking like this.  All test cases are passing - 

```
        if not args and not kwargs:
            raise ValueError("You must provide either a UUID and data dict or a list of objects to update")
        uuid = kwargs.get("id", "")
        data = kwargs.get("data", {})
        if data and not uuid:
            uuid = args[0]
        if len(args) == 2:
            uuid, data = args
        
        if not uuid and not data:
            objects = args[0] if args else []
            bulk_data = []

            if not isinstance(objects, list):
                raise ValueError("objects must be a list[dict()|Record] not " + str(type(objects)))

            for o in objects:
                try:
                    if isinstance(o, dict):
                        bulk_data.append(o)
                    elif isinstance(o, Record):
                        if not hasattr(o, "id"):
                            raise ValueError("'Record' object has no attribute 'id'")
                        updates = o.updates()
                        if updates:
                            updates["id"] = o.id
                            bulk_data.append(updates)
                    else:
                        raise ValueError("Invalid object type: " + str(type(o)))
                except ValueError as exc:
                    raise ValueError("Unexpected value in object list") from exc

        req = Request(
            key=uuid,
            base=self.url,
            token=self.api.token,
            api_version=self.api.api_version,
        )
        if uuid and data:
            if req.patch(data):
                return True
            return False
        if objects and bulk_data:
            return response_loader(req.patch(bulk_data), self.return_obj, self)
```

joewesch: I have a better idea, let's add the bulk update method and call it via the update method.
```python
    def update(self, *args, **kwargs):
        # Original signature: update(self, id: str, data: Dict[str, any])
        if not args and not kwargs:
            raise ValueError("You must provide either a UUID and data dict or a list of objects to update")
        uuid = kwargs.get("id", "")
        data = kwargs.get("data", {})
        if data and not uuid:
            uuid = args[0]
        if len(args) == 2:
            uuid, data = args

        if not any([uuid, data]):
            return self.bulk_update(args[0])

        req = Request(
            key=uuid,
            base=self.url,
            token=self.api.token,
            api_version=self.api.api_version,
        )
        if req.patch(data):
            return True
        return False

    def bulk_update(self, objects: List[Dict[str, Any]]):
        if not isinstance(objects, list):
            raise ValueError("objects must be a list[dict()|Record] not " + str(type(objects)))

        bulk_data = []
        for o in objects:
            try:
                if isinstance(o, dict):
                    bulk_data.append(o)
                elif isinstance(o, Record):
                    if not hasattr(o, "id"):
                        raise ValueError("'Record' object has no attribute 'id'")
                    updates = o.updates()
                    if updates:
                        updates["id"] = o.id
                        bulk_data.append(updates)
                else:
                    raise ValueError("Invalid object type: " + str(type(o)))
            except ValueError as exc:
                raise ValueError("Unexpected value in object list") from exc

        req = Request(
            base=self.url,
            token=self.api.token,
            api_version=self.api.api_version,
        ).patch(bulk_data)
        return response_loader(req, self.return_obj, self)
```

vu-smitt13: I like this approach much better!

joewesch (approved):
Looks great, thanks!

PR #171, opened 2024-03-19
add support for setting a HTTP user agent

teunvink:
This PR adds support for specifying a HTTP user agent string. 

Rationale: in our codebase we try to set user agent strings to the script calling the API. This makes it easier to track down which scripts on which machines use specific APIs, which can become quite difficult to figure out on machines behind NAT and with many different scripts and users on them.

### DIFF
--- a/pynautobot/core/api.py
+++ b/pynautobot/core/api.py
@@ -62,7 +62,7 @@ class Api(object):
     :param int,optional retries: Number of retries, for HTTP codes 429, 500, 502, 503, 504,
         this client will try before dropping.
     :param bool,optional verify: SSL cert verification.
-    :param dict,optional headers: Set add additional HTTP headers
+    :param dict,optional headers: Override default or add additional HTTP headers
     :raises AttributeError: If app doesn't exist.
     :Examples:
 
--- a/pynautobot/core/api.py
+++ b/pynautobot/core/api.py
@@ -62,6 +62,7 @@ class Api(object):
     :param int,optional retries: Number of retries, for HTTP codes 429, 500, 502, 503, 504,
         this client will try before dropping.
     :param bool,optional verify: SSL cert verification.
+    :param str,optional http_user_agent: Override the default User-Agent used for HTTP requests
     :raises AttributeError: If app doesn't exist.
     :Examples:
 
@@ -82,6 +83,7 @@ class Api(object):
         api_version=None,
         retries=0,
         verify=True,
+        http_user_agent=None,
     ):
         base_url = "{}/api".format(url if url[-1] != "/" else url[:-1])
         self.token = token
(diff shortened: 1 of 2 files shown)


joewesch:
Setting some headers is also talked about in #158. I don't mind accepting different headers, but I don't want to have to add a param for every header possible. Instead of what you have here, I think accepting a `headers` param is better and then having the input headers override the default headers.

```python
self.headers = {"Authorization": f"Token {self.token}"}
if headers:
    self.headers.update(**headers)
```

teunvink:
@joewesch yes, that makes sense. I hadn't seen that isssue yet, I probably only searched for 'User-Agent'.  Should I close this PR, update it with your suggestion?

Adding it just to `self.headers` won't be enough though (as far as I could see in my tests), you need to add it to the headers in `self.http_session`.

joewesch:
> Should I close this PR, update it with your suggestion?

No, that was a WIP PR and they haven't updated it in a while. If you get this one updated we can close that one and tell them to use the pattern we set here instead.

> Adding it just to self.headers won't be enough though (as far as I could see in my tests), you need to add it to the headers in self.http_session.

That's fine, it was merely a mock up. Add it where it makes sense (or both).

pynautobot/core/api.py
```diff
@@ -103,6 +107,8 @@ def __init__(
         self.threading = threading
         self.max_workers = max_workers
         self.api_version = api_version
+        if headers:
+            self.http_session.headers.update(**headers)
```
joewesch, line 111: Let's combine and move up with the rest of the `self.http_session` items:
```python
        self.http_session = requests.Session()
        self.http_session.verify = verify
        if headers:
            self.headers.update(**headers)
            self.http_session.headers.update(**headers)
```

pynautobot/core/api.py
```diff
@@ -62,6 +62,7 @@ class Api(object):
     :param int,optional retries: Number of retries, for HTTP codes 429, 500, 502, 503, 504,
         this client will try before dropping.
     :param bool,optional verify: SSL cert verification.
+    :param dict,optional headers: Set add additional HTTP headers
```
joewesch, line 65:
```suggestion
    :param dict,optional headers: Override default or add additional HTTP headers
```

joewesch:
Let's also update this section of the docs:
https://github.com/nautobot/pynautobot/blob/66ebf7f3117b731661510c6dd78f9f80bbbe4291/docs/advanced/session.rst?plain=1#L13-L31

teunvink:
@joewesch just to be sure, do I need to do anything more with this, or is this just awaiting approval?

teunvink:
Ah, I misunderstood that. I've updated the docs, hopefully this is OK.

joewesch:
Ah, it looks like you may need to rebase from develop again and instead update `docs/dev/advanced/session.md` as the docs have been migrated to markdown. Sorry for run around.

teunvink:
Reading this documentation (which I seem to have missed before) makes me realise my PR isn't needed. This works just fine:
```
nautobot.http_session.headers["User-Agent"] = "my-user-agent"
```
I'm closing this PR, sorry for the noise.

Issue #172, opened 2024-03-27
RequestError: Authentication credentials were not provided


joewesch:
Details are in https://github.com/nautobot/nautobot/issues/5490

PR #174, opened 2024-03-27
Fixes pynautobot for v1.6.16 auth changes

joewesch:
v1 fix for #172 

This PR adds the token to status, version and openapi calls. I also updated our CI to only test nautobot v1.6 as well as added python 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,11 +10,8 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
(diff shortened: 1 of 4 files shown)


jvanderaa (approved):

jvanderaa:
Looks good. We should open an issue to make a new release with requiring the token as part of the initialization as well as an enhancement. Not holding this back though.

joewesch:
> We should open an issue to make a new release with requiring the token as part of the initialization as well as an enhancement.

I believe the reason why it's not a required argument is because you can also use the environment variable. It is still ultimately "required" though to do almost any function.

PR #180, opened 2024-04-11
Updates requests library and removes urllib pin

jvanderaa:
Removes a pinned dependency for urllib and updates requests to latest.


glennmatthews:
I'm guessing it wasn't intentional to also re-lock dependencies like `black` to new versions?

jvanderaa:
You are correct that it was unintentional. I think I need to do a poetry remove on urllib and poetry update requests with those two. Will fix up.

PR #184, opened 2024-04-27
WIP: Migrates initial set to mkdocs.

jvanderaa:
This is the work in progress towards getting the pynautobot docs to use mkdocs as the platform.

Outstanding

- [ ] Advanced Update
- [ ] Attempt to auto-generate some API docs pages
- [ ] Review existing docs over to the method


mkdocs.yml
```diff
@@ -0,0 +1,170 @@
+---
+# TODO
+# Change dev address
+dev_addr: "0.0.0.0:8002"
```
joewesch, line 4: Is this TODO still valid?

pyproject.toml
```diff
@@ -43,6 +43,17 @@ Sphinx = "*"
 [tool.poetry.extras]
 docs = ["sphinx"]
 
+[tool.poetry.group.dev.dependencies]
```
joewesch, line 46: Do we need both this group and the `[tool.poetry.dev-dependencies]` group?

jvanderaa: No, I'd like to move the existing one into this, but it breaks things. Thoughts?

mkdocs.yml
jvanderaa: Yes. Fixing.

pyproject.toml
glennmatthews: I'd try `poetry remove --dev <package>` followed by `poetry add --group dev <package>` as the migration process.

mkdocs.yml
```diff
@@ -0,0 +1,170 @@
...
+      - Library Overview: "user/lib_overview.md"
+  - Adminsitrator Guide:
+      - Install and Configure: "admin/install.md"
+      - Upgrade: "admin/upgrade.md"
+      - Uninstall: "admin/uninstall.md"
+      - Release Notes:
+          - "admin/release_notes/index.md"
+          - v2.1: "admin/release_notes/version_2.1.md"
+          - v2.0: "admin/release_notes/version_2.0.md"
+          - v2.0-rc1: "admin/release_notes/version_2.0-rc1.md"
```
cmsirbu, line 116: We don't generally add release notes for alpha/beta/rc etc. I would remove this, or if you think there's any value in keeping the notes, move them within the `v2.0` page since they belong to that version train.

joewesch (dismissed):

joewesch (approved):

PR #193, opened 2024-05-08
Moves and updates CSS.

jvanderaa:
Moves basic usage into User Guide, which makes more sense for its location.

Updates CSS styling so there is some padding on a border that would have otherwise been overlapping.


docs/assets/extra.css
```diff
@@ -147,4 +147,17 @@ img.copyright-logo {
...
+    padding-left: 0.5rem;
+    padding-right: 0.5rem;
+    padding-top: 0.5rem;
+    padding-bottom: 0.5rem;
+}
+
+.md-typeset .admonition,
+.md-typeset details {
+    border-width: 0;
+    border-left-width: 4px;
```
tsm1th, line 162: Was the intent to remove the border on these elements so that it's only viewable on the left side?

image

After:

image

jvanderaa: Yes it is. Looks a bit better on non-dark mode @tsm1th

glennmatthews: This should probably go through dev-standards and the app cookiecutter - we should have consistent styling across all of our docs?

jvanderaa: I like the idea.

jvanderaa: Removed this from this PR

joewesch (approved):
LGTM

PR #203, opened 2024-06-21
fix(Record): Cast Record.__str__ return to str

jamesharr:
## New Pull Request

Have you:
- [ ] Updated the README if necessary?
- [ ] Updated any configuration settings?
- [ ] Written a unit test?

## Justification

Cast Record.__str__ to string before returning.

For us, this fixes cases where `name` on a custom object is an integer and repr(obj) is called.

### DIFF
--- a/pynautobot/core/response.py
+++ b/pynautobot/core/response.py
@@ -202,7 +202,7 @@ class Record(object):
         return dict(self)[k]
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


jamesharr:
Let me know if you'd like me to do anything for this (like tests)

joewesch (approved):

PR #206, opened 2024-07-03
Set the user-agent to python-pynautobot/__version__ by default (ltm-1.6)

nrnvgh:
Added code and a test case for setting the user-agent to a default value.

### DIFF
--- a/pynautobot/core/api.py
+++ b/pynautobot/core/api.py
@@ -21,7 +21,6 @@ import requests
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
-import pynautobot
 from pynautobot.core.query import Request
 from pynautobot.core.app import App, PluginsApp
 from pynautobot.core.graphql import GraphQLQuery
@@ -85,13 +84,15 @@ class Api(object):
         retries=0,
         verify=True,
     ):
+        from pynautobot import __version__
+
         base_url = "{}/api".format(url if url[-1] != "/" else url[:-1])
         self.token = token
         self.headers = {"Authorization": f"Token {self.token}"}
         self.base_url = base_url
         self.http_session = requests.Session()
         self.http_session.verify = verify
-        self.http_session.headers.update({"User-Agent": f"python-pynautobot/{pynautobot.__version__}"})
+        self.http_session.headers.update({"User-Agent": f"python-pynautobot/{__version__}"})
         if retries:
             _adapter = HTTPAdapter(
                 max_retries=Retry(
--- a/pynautobot/core/api.py
+++ b/pynautobot/core/api.py
@@ -21,6 +21,7 @@ import requests
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
+import pynautobot
 from pynautobot.core.query import Request
 from pynautobot.core.app import App, PluginsApp
 from pynautobot.core.graphql import GraphQLQuery
@@ -90,6 +91,7 @@ class Api(object):
         self.base_url = base_url
         self.http_session = requests.Session()
         self.http_session.verify = verify
+        self.http_session.headers.update({"User-Agent": f"python-pynautobot/{pynautobot.__version__}"})
         if retries:
             _adapter = HTTPAdapter(
                 max_retries=Retry(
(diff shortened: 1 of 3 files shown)


pynautobot/core/api.py
```diff
@@ -90,6 +91,7 @@ def __init__(
         self.base_url = base_url
         self.http_session = requests.Session()
         self.http_session.verify = verify
+        self.http_session.headers.update({"User-Agent": f"python-pynautobot/{pynautobot.__version__}"})
```
joewesch, line 94: Should we instead add it to `self.headers` above and do `self.http_session.headers.update(**self.headers)` here? It looks like the only place we currently use `self.headers` is in [the graphql query](https://github.com/nautobot/pynautobot/blob/0473056b7564632d9e01ad5fdca17fe12453e1f7/pynautobot/core/graphql.py#L126).

nrnvgh: Can, but that's potentially more invasive; I took the path I did because I wanted something with as few changes as possible for acceptance into `ltm-1.6`. I'm happy to do as you suggest, but since `self.http_session.headers.update(**self.headers)` means that `pynautobot.core.api.Api.__init__()` would be setting an`Authorization` header, should I also remove the additions of the `Authorization` header in:
-  `pynautobot.core.query.Request.get_openapi()`
-  `pynautobot.core.query.Request.get_version()`
- `pynautobot.core.query.Request.get_status()`

?

Not strictly necessary, but it feels cleaner to me to do it that way. And `Authorization` is also being set in `pynautobot.core.query._make_call()` and being passed by most (all?) calls to `Request()`. If we add the `Authorization` header to `self.http_session.headers`, then shouldn't the calls to `Request()` be modified to no longer pass a `token` value? Arguably, `Request` could be modified to no longer /accept/ in, but that could totally go either way.

None of this cleanup is required for the functionality provided by this patch, but it feels like a question which should be answered (if not right now, then at some point). I wouldn't be surprised if I've missed something, but this is where my head's at; thoughts?

joewesch: As with most collaborative code bases, the methods that we use for different pieces of code that carry out similar functions can obviously vary based on many factors such as haste, copy pasta, inexperience or insecurity. The addition of the Authorization header to the `get_openapi`, `get_version` and `get_status` endpoints were done as a quick break-fix PR (#174) that was more "fix it ASAP" rather than a thought out design decision. You can clearly see [one of them was completely redundant](https://github.com/nautobot/pynautobot/blob/c1c2a8f2a8b8c2a5093c0b69a6d289787b160514/pynautobot/core/query.py#L223-L226)!

So, I apologize that I opened that can of worms and I relent that the simplest solution is probably the best for now.

joewesch (approved):

nrnvgh:
Checking in on this; I'm don't need this feature Right This Second or anything, I just want to be sure getting it merged in is still on the radar.

joewesch:
Yup, I just don't like to be hasty when it comes to LTM changes. I have asked @jvanderaa to review as well.

nrnvgh:
Totally get it; cheers.

jvanderaa:
I don't feel great about importing pynautobot from pynautobot. Can we do something with a property or something? @joewesch

pynautobot/core/api.py
```diff
@@ -21,6 +21,7 @@
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
+import pynautobot
```
joewesch, line 24: We have a fear of possibly causing circular imports in the future, so please change this import to happen inside the `__init__` method and change it to only import `__version__`:

```python
    def __init__(
        ...
    ):
        from pynautobot import __version__  # pylint: disable=import-outside-toplevel

        base_url = "{}/api".format(url if url[-1] != "/" else url[:-1])
        ...
        self.http_session.headers.update({"User-Agent": f"python-pynautobot/{__version__}"})
```

nrnvgh: Not a problem; update committed to both branches.

jvanderaa (approved):

joewesch (approved):

PR #207, opened 2024-07-03
Set the user-agent to python-pynautobot/__version__ by default (develop)

nrnvgh:
Added code and a test case for setting the user-agent to a default value.


joewesch:
Let's keep discussions in your LTM PR (#206) and wait until that one is approved and merged before we can replicate any changes and complete this one.

joewesch (approved):

PR #216, opened 2024-07-17
Added integration test that would close #102

### LINKED ISSUE #102: Add integration test for jobs run method
Unittest are included, but writing a integration test will provide additional coverage and validation.

tsm1th:
This integration test would close #102.


tsm1th:
@jeffkala - I don't know if this is feasible or not.  The CI pipeline keeps failing for reasons mentioned [here](https://github.com/orgs/nautobot/discussions/5553).  I could see this being an intermittent failure to the pipeline going forward with it's dependency on the celery worker.  Thoughts?

jvanderaa:
What if we moved them all to requests mock?

joewesch:
> What if we moved them all to requests mock?

I think you're thinking of a unit test. The request was for an integration test.

PR #217, opened 2024-07-18
add limit and offset features

### LINKED ISSUE #106: Allow for manual pagination

tsm1th:
This is the start of the work to close #106 

- [x] Add unit test
- [x] Add integration test
- [x] Add examples in documentation

### DIFF
--- a/pynautobot/core/query.py
+++ b/pynautobot/core/query.py
@@ -22,9 +22,9 @@ import json
 import requests
 
 
-def calc_pages(limit, count, offset):
+def calc_pages(limit, count):
     """Calculate number of pages required for full results set."""
-    return int((count - offset) / limit) + (limit % count > 0)
+    return int(count / limit) + (limit % count > 0)
 
 
 class RequestError(Exception):
@@ -327,7 +327,7 @@ class Request(object):
             req = self._make_call(add_params=add_params)
             if isinstance(req, dict) and req.get("results") is not None:
                 ret = req["results"]
-                while req["next"]:
+                while req["next"] and self.offset is None:
                     if not add_params:
                         req = self._make_call(add_params={"limit": req["count"], "offset": len(req["results"])})
                     else:
@@ -344,10 +344,10 @@ class Request(object):
             req = self._make_call(add_params=add_params)
             if isinstance(req, dict) and req.get("results") is not None:
                 ret = req["results"]
-                if req.get("next"):
+                if req.get("next") and self.offset is None:
                     page_size = len(req["results"])
-                    pages = calc_pages(page_size, req["count"], self.offset or 0)
-                    page_offsets = [((increment * page_size) + (self.offset or 0)) for increment in range(1, pages)]
+                    pages = calc_pages(page_size, req["count"])
+                    page_offsets = [increment * page_size for increment in range(1, pages)]
                     if pages == 1:
                         req = self._make_call(url_override=req.get("next"))
                         ret.extend(req["results"])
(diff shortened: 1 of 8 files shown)


pynautobot/core/query.py
```diff
@@ -22,9 +22,10 @@
 import requests
 
 
-def calc_pages(limit, count):
+def calc_pages(limit, count, offset):
     """Calculate number of pages required for full results set."""
-    return int(count / limit) + (limit % count > 0)
+    print(limit, count, offset)
```
joewesch, line 27:
```suggestion
```

I'm assuming this was leftover.

tsm1th: Maybe!!!! :)

tsm1th:
Thinking more about this.  If a user provided "offset", then should we assume they only want that single chunk of data?

Example - with some dataset containing 100 records

If I want offset=10 and limit=10, then do I only want records 10-20?  Or do I want records 10-100?

Without an offset, I think it's obvious that I want all the records, but chunked to the selected limit.  

With an offset, I think it might be beneficial to only return that single chunk... (not the current logic)

Thoughts?

joewesch:
Hmm, good question. Maybe we need `page_size` to signify that they just want to change the default paging and indeed return only the specific chunk when they use limit/offset? Since this was @Kircheneer's request I'd appreciate his feedback and thoughts.

tsm1th:
And for clarity, limit equals page_size in the Nautobot [API](https://docs.nautobot.com/projects/core/en/stable/user-guide/administration/configuration/optional-settings/#max_page_size)

joewesch:
Yes, I just meant a way to differentiate. So you're saying if they just put `limit` they would modify the page size, but if they provide both `limit` and `offset` we should treat it as a single chunk wanted instead? I'm fine with that logic.

tsm1th:
@joewesch - The SDK is doing a bit of magic for the users that wouldn't exist if they were just using the API.  For example, `limit` is just controlling the page size like you said, but we're retrieving all the results for them instead of making the user do  it.  The question is whether or not we do that same magic when `limit` and `offset` are both defined.  (It currently is returning all the chunks for both)

I think it could be beneficial to provide that `single chunk` functionality for someone using the SDK who might want to handle the chunking themselves.  But, why not just use the API if you're needing to go as far as managing the chunking?

joewesch:
> The question is whether or not we do that same magic when limit and offset are both defined. (It currently is returning all the chunks for both)

If we are going to add the options then it should only return a single chunk if they provide both and only change the page size if they only provide `limit`. Only providing `offset` should raise the exception as you already have.

> I think it could be beneficial to provide that single chunk functionality for someone using the SDK who might want to handle the chunking themselves. But, why not just use the API if you're needing to go as far as managing the chunking?

I can see a few edge cases that it would come in handy. For example, as a replacement for a (currently non-existent) `.first()` convenience method. Again this is Leo's request so I would like him to weigh in.

Kircheneer:
I don't actually need this anymore and don't really have anything to contribute, sorry.

joewesch (approved):

PR #224, opened 2024-07-26
Fixed issue where circuit details are not known without fetching them

### LINKED ISSUE #222: Circuits has no attribute cid
I found a bug by trying to enumerate the cables in the Nautobot demo. Here is the code to get the bug.

```python
import pynautobot

nautobot = pynautobot.api(
    url="https://demo.nautobot.com",
    token="aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
)

for cable in nautobot.dcim.cables.all():
    print(cable)
```

I was expecting to get a string representation of all the cables but instead, I got the following exception.

```
Ethernet1/1 <> Ethernet1/1
  File "c:\Users\foo\Documents\misc\circuit_bug.py", line 14, in <module>
    print(cable)
  File "C:\Users\foo\AppData\Roaming\Python\Python310\site-packages\pynautobot\models\dcim.py", line 228, in __str__
    return "{} <> {}".format(self.termination_a, self.termination_b)
  File "C:\Users\foo\AppData\Roaming\Python\Python310\site-packages\pynautobot\models\dcim.py", line 217, in __str__
    return self.circuit.cid
AttributeError: type object 'Circuits' has no attribute 'cid'
```

I think it is related to the termination type, it seems not to like circuits. 

Maybe the `__str__` function of the `Termination` class is a bit too hacky (cf. [model/dcim.py](https://github.com/nautobot/pynautobot/blob/89793438fd8ac59a225ebd18938b503835187384/pynautobot/models/dcim.py)).

Nautobot version: 2.2.8
Pynautobot version: 2.2.0

tsm1th:
This closes #222.  

The other termination type includes the identifier `name` in the original API call, so it's not an issue.  For a circuit termination type, the identifier `circuit.cid` is not in the original API call.  That's why an additional fetch is necessary.  

Pynautobot usually handles this with the `__getattr__` method.  However, this does not work in this case because the `Circuit` object has not been initialized yet.   Once initialized via `full_details()`, the `__getattr__` works as expected.  

Another solution was to increase the `depth` param on the API call, but it added performance overhead that wasn't ideal.

### DIFF
--- a/pynautobot/models/dcim.py
+++ b/pynautobot/models/dcim.py
@@ -213,16 +213,12 @@ class Termination(Record):
         # hacky check to see if we're a circuit termination to
         # avoid another call to Nautobot because of a non-existent attr
         # in self.name
-        # Fetch details for the circuit if it doesn't have them
         if "circuit" in str(self.url):
-            if not self.has_details:
-                self.full_details()
             return self.circuit.cid
 
         return self.name
 
     device = Devices
-    circuit = Circuits
 
 
 class Cables(Record):
--- a/pynautobot/models/dcim.py
+++ b/pynautobot/models/dcim.py
@@ -213,7 +213,10 @@ class Termination(Record):
         # hacky check to see if we're a circuit termination to
         # avoid another call to Nautobot because of a non-existent attr
         # in self.name
+        # Fetch details for the circuit if it doesn't have them
         if "circuit" in str(self.url):
+            if not self.has_details:
+                self.full_details()
             return self.circuit.cid
 
         return self.name
--- a/pynautobot/models/dcim.py
+++ b/pynautobot/models/dcim.py
@@ -19,7 +19,6 @@ from requests.utils import urlparse
 from pynautobot.core.query import Request
 from pynautobot.core.response import Record, JsonField
 from pynautobot.core.endpoint import RODetailEndpoint
-from pynautobot.models.circuits import Circuits
 
 
 class TraceableRecord(Record):
--- a/pynautobot/models/dcim.py
+++ b/pynautobot/models/dcim.py
@@ -19,6 +19,7 @@ from requests.utils import urlparse
 from pynautobot.core.query import Request
 from pynautobot.core.response import Record, JsonField
 from pynautobot.core.endpoint import RODetailEndpoint
+from pynautobot.models.circuits import Circuits
 
 
 class TraceableRecord(Record):
@@ -209,15 +210,10 @@ class Racks(Record):
 
 class Termination(Record):
     def __str__(self):
-        # hacky check to see if we're a circuit termination to
-        # avoid another call to Nautobot because of a non-existent attr
-        # in self.name
-        if "circuit" in str(self.url):
-            return self.circuit.cid
-
-        return self.name
+        return self.display
 
     device = Devices
+    circuit = Circuits
 
 
 class Cables(Record):
--- a/tests/test_dcim.py
+++ b/tests/test_dcim.py
@@ -354,7 +354,7 @@ class CablesTestCase(Generic.Tests):
                         "url": "http://localhost:8000/api/circuits/circuits/1/",
                     },
                     "term_side": "A",
-                    "display": "TEST123321"
+                    "display": "TEST123321",
                 },
                 "termination_b_type": "dcim.interface",
                 "termination_b_id": 2,
--- a/tests/test_dcim.py
+++ b/tests/test_dcim.py
@@ -340,7 +340,7 @@ class CablesTestCase(Generic.Tests):
     app = "dcim"
     name = "cables"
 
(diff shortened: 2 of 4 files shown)


pynautobot/models/dcim.py
```diff
@@ -213,7 +213,10 @@ def __str__(self):
         # hacky check to see if we're a circuit termination to
         # avoid another call to Nautobot because of a non-existent attr
         # in self.name
+        # Fetch details for the circuit if it doesn't have them
         if "circuit" in str(self.url):
+            if not self.has_details:
```
jvanderaa, line 218: Would it make sense here to move to a property?

tsm1th: Not sure I understand.  Move `has_details` to a property?

tsm1th: I think I get it or if not then it sparked another idea.  We can add `circuit` as a property to the `Termination` model I believe and that might be the better solution here.

joewesch:
Can you add a test to cover what this actually fixes?

tsm1th:
@joewesch - The solution here was much simpler than I realized.  The original issue stemmed from trying to rely on data from the termination object's parent to generate a `__str__` method.  This is not how nautobot does it.  The termination object itself has a `display` attribute that handles this.  This is viewable in the api when you look at a `dcim.interface`, `circuits.circuittermination`, or other termination object type.

TLDR - There was no reason to generate our own `__str__` method using the parent of the termination object.

I updated the test that was checking this (to include the `display` attribute) and renamed it to more accurately demonstrate what it's doing.

tests/test_dcim.py
```diff
@@ -352,9 +352,9 @@ def test_get_circuit(self):
                     "circuit": {
                         "id": 346,
                         "url": "http://localhost:8000/api/circuits/circuits/1/",
-                        "cid": "TEST123321",
```
joewesch, line 355: Did this line need to be removed?

tsm1th: Not necessarily.  That's an accurate model of the data if the `depth > 0`.  I can adjust all the responses if we want to mimic the real response that pynautobot gets.

tsm1th: An example from the api explorer.  You can see that `cid` does not actually exist.  

```json 
{
  "id": "e2076809-21cd-437e-a9bd-0c7ae9f9164f",
  "object_type": "circuits.circuittermination",
  "display": "Termination A: ORD02",
  "url": "https://demo.nautobot.com/api/circuits/circuit-terminations/e2076809-21cd-437e-a9bd-0c7ae9f9164f/",
  "natural_slug": "telia-carrier_tel-90526601020494853_a_e207",
  "cable_peer_type": "dcim.interface",
  "cable_peer": {
    "id": "0105c0cc-a47f-4f89-9f75-c9344aa3b2e6",
    "object_type": "dcim.interface",
    "url": "https://demo.nautobot.com/api/dcim/interfaces/0105c0cc-a47f-4f89-9f75-c9344aa3b2e6/"
  },
  "connected_endpoint_type": "dcim.interface",
  "connected_endpoint": {
    "id": "0105c0cc-a47f-4f89-9f75-c9344aa3b2e6",
    "object_type": "dcim.interface",
    "url": "https://demo.nautobot.com/api/dcim/interfaces/0105c0cc-a47f-4f89-9f75-c9344aa3b2e6/"
  },
  "connected_endpoint_reachable": true,
  "term_side": "A",
  "port_speed": null,
  "upstream_speed": null,
  "xconnect_id": "",
  "pp_info": "",
  "description": "",
  "cable": {
    "id": "cade60f1-2e7f-476e-b765-fea43ce80bcc",
    "object_type": "circuits.circuittermination",
    "url": "https://demo.nautobot.com/api/dcim/cables/cade60f1-2e7f-476e-b765-fea43ce80bcc/"
  },
  "circuit": {
    "id": "72998f3b-376f-466a-b4e6-617c09a69376",
    "object_type": "circuits.circuit",
    "url": "https://demo.nautobot.com/api/circuits/circuits/72998f3b-376f-466a-b4e6-617c09a69376/"
  },
  "location": {
    "id": "bac30ffb-18e5-472f-a067-7e5fbeb2d816",
    "object_type": "dcim.location",
    "url": "https://demo.nautobot.com/api/dcim/locations/bac30ffb-18e5-472f-a067-7e5fbeb2d816/"
  },
  "provider_network": null,
  "created": "2023-09-21T00:00:00Z",
  "last_updated": "2023-09-21T19:49:52.064156Z",
  "notes_url": "https://demo.nautobot.com/api/circuits/circuit-terminations/e2076809-21cd-437e-a9bd-0c7ae9f9164f/notes/",
  "custom_fields": {}
}
```

joewesch (approved):

PR #233, opened 2024-09-03
Added graphql run method for query objects

### LINKED ISSUE #232: Execute saved graphql queries via pynautobot
It would be nice to be able to execute saved queries stored in nautobot directly through pynautobot including supplying any additional variable definitions the query my require.

tsm1th:
This would close #232 if approved.  

Usage:
```python
query = nb.extras.graphql_queries.get("example")
query.run(
    variables={"foo": "bar"})
)
```

### DIFF
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -695,7 +695,7 @@ class JobsEndpoint(Endpoint):
 class GraphqlEndpoint(Endpoint):
     """Extend Endpoint class to support run method for graphql queries."""
 
-    def run(self, *args, query_id=None, **kwargs):
+    def run(self, query_id, *args, **kwargs):
         """Runs a saved graphql query based on the query_id provided.
 
         Takes a kwarg of `query_id` to specify the query that should be run.
@@ -728,11 +728,7 @@ class GraphqlEndpoint(Endpoint):
                     {"variables":{"foo":"bar"}}
                 )
         """
-
-        if not query_id:
-            raise ValueError('Keyword Argument "query_id" is required to run a query.')
         query_run_url = f"{self.url}/{query_id}/run/"
-
         return Request(
             base=query_run_url,
             token=self.token,
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -690,3 +690,39 @@ class JobsEndpoint(Endpoint):
         ).post(args[0] if args else kwargs)
 
         return response_loader(req, self.return_obj, self)
+
+
+class GraphqlEndpoint(Endpoint):
+    """Extend Endpoint class to support run method for graphql queries."""
+
+    def run(self, *args, **kwargs):
+        """Runs a saved graphql query based on the query_id provided.
+
+        Takes a kwarg of `query_id` to specify the query that should be run.
+
+        Args:
+            *args (str, optional): Used as payload for POST method
+                to the API if provided.
+            **kwargs (str, optional): Any search argument the
+                endpoint accepts can be added as a keyword arg.
+
+        Returns:
+            An API response from the execution of the saved graphql query.
+
+        Examples:
+            To run a query with the following variables:
+            >>> query = nb.extras.graphql_queries.get("Example")
+            >>> query.run(
+                    variables={"foo": "bar"})
+                )
+        """
+
+        if not kwargs.get("query_id"):
+            raise ValueError('Keyword Argument "query_id" is required to run a query.')
+        query_run_url = f"{self.url}/{kwargs['query_id']}/run/"
+
+        return Request(
+            base=query_run_url,
+            token=self.token,
+            http_session=self.api.http_session,
+        ).post(args[0] if args else kwargs)
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -695,7 +695,7 @@ class JobsEndpoint(Endpoint):
 class GraphqlEndpoint(Endpoint):
     """Extend Endpoint class to support run method for graphql queries."""
 
-    def run(self, *args, **kwargs):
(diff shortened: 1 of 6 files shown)


pynautobot/core/endpoint.py
```diff
@@ -690,3 +690,39 @@ def run(self, *args, api_version=None, **kwargs):
...
+
+        if not kwargs.get("query_id"):
+            raise ValueError('Keyword Argument "query_id" is required to run a query.')
+        query_run_url = f"{self.url}/{kwargs['query_id']}/run/"
+
+        return Request(
+            base=query_run_url,
+            token=self.token,
+            http_session=self.api.http_session,
+        ).post(args[0] if args else kwargs)
```
joewesch, line 728: What happens if you run a query without any required variables? Is the error message descriptive enough to pass through to the end user or should we catch the error and reword it?

joewesch:
Let's add something to the docs as well, please.

pynautobot/core/endpoint.py
```diff
@@ -690,3 +690,39 @@ def run(self, *args, api_version=None, **kwargs):
...
+        Examples:
+            To run a query with the following variables:
+            >>> query = nb.extras.graphql_queries.get("Example")
+            >>> query.run(
+                    variables={"foo": "bar"})
+                )
+        """
+
+        if not kwargs.get("query_id"):
+            raise ValueError('Keyword Argument "query_id" is required to run a query.')
```
jmcgill298, line 721: Why not make `query_id` a required arg?

pynautobot/core/endpoint.py
```diff
@@ -690,3 +690,39 @@ def run(self, *args, api_version=None, **kwargs):
         ).post(args[0] if args else kwargs)
 
         return response_loader(req, self.return_obj, self)
+
+
+class GraphqlEndpoint(Endpoint):
```
joewesch, line 695: Do we need to add this new endpoint here: https://github.com/nautobot/pynautobot/blob/develop/pynautobot/core/app.py#L63-L66

pynautobot/core/endpoint.py
tsm1th: It's currently caught with the `query.RequestError` exception, but the API message is fairly descriptive:
```
  File "/home/ntc/pynautobot/connect.py", line 26, in <module>
    print(variable_query.run())
  File "/home/ntc/pynautobot/pynautobot/models/extras.py", line 50, in run
    return GraphqlEndpoint(self.api, self.api.extras, "graphql_queries").run(query_id=self.id, **kwargs)
  File "/home/ntc/pynautobot/pynautobot/core/endpoint.py", line 728, in run
    ).post(args[0] if args else kwargs)
  File "/home/ntc/pynautobot/pynautobot/core/query.py", line 398, in post
    return self._make_call(verb="post", data=data)
  File "/home/ntc/pynautobot/pynautobot/core/query.py", line 294, in _make_call
    raise RequestError(req)
pynautobot.core.query.RequestError: The request failed with code 400 Bad Request: {'errors': [{'message': 'Variable "$devicename" of required type "[String]!" was not provided.', 'locations': [{'line': 1, 'column': 15}]}]}
```

pynautobot/core/endpoint.py
tsm1th: Not in my testing, but not sure of everything added for the `jobs` integration.  Is it possible that `jobs` can be a child object of something else?  I don't think the same is true for a saved query.

pynautobot/core/endpoint.py
joewesch: Ok, that is fine. Thanks!

pynautobot/core/endpoint.py
joewesch: It looks like it's required to be able to call the run method directly with `nb.extras.graphql_queries.run()`:

```python
In [11]: nb.extras.jobs.run
Out[11]: <bound method JobsEndpoint.run of <pynautobot.core.endpoint.JobsEndpoint object at 0x7f9b38d05d00>>

In [12]: nb.extras.graphql_queries.run
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-12-bc5064448ad1> in <cell line: 1>()
----> 1 nb.extras.graphql_queries.run

AttributeError: 'Endpoint' object has no attribute 'run'
```

Does it make sense to call that directly like that? I'm not sure if that's necessarily a requirement, but maybe we should add it for consistency if it does.

pynautobot/core/endpoint.py
jmcgill298: @tsm1th I don't think you understood what I was saying. I meant doing `def run(query_id, *args, **kwargs):` and letting the standard invalid arg error from python be raised, and thus we don't need these two lines.

pynautobot/core/endpoint.py
tsm1th: Agreed, added it for consistency.

pynautobot/core/endpoint.py
tsm1th: I could make it a positional argument, but I would have to change the `*args` logic or remove it.  I'm currently allowing a custom payload to be passed in via args.

Making that change results in the following new error instead:
`TypeError: GraphqlEndpoint.run() missing 1 required positional argument: 'query_id'`

tsm1th: This would work:

```python
class GraphqlQueries(Record):
    def run(self, *args, **kwargs):
        """Run a graphql query from a saved graphql instance."""
        return GraphqlEndpoint(self.api, self.api.extras, "graphql_queries").run(self.id, *args, **kwargs)
```

And in the endpoint:
```python
class GraphqlEndpoint(Endpoint):
    """Extend Endpoint class to support run method for graphql queries."""

    def run(self, query_id, *args, **kwargs):
```

joewesch (approved):

PR #240, opened 2024-10-09
Changed all() method to call filter() method in Endpoint class

### LINKED ISSUE #237: Support include kwarg in the Endpoint.all() Method
Today if you want to pull back all devices and include config_context data you must use the `.filter` call.

In the .all() call if you include the `include` kwarg it will fail

```
>>> nautobot.dcim.devices.all(include=['config_context'])
  File "<stdin>", line 1, in <module>
TypeError: Endpoint.all() got an unexpected keyword argument 'include'
>>> 
```

Workaround is to use `.filter(include=["config_context"])` with no other filtering criteria.

```
>>> nautobot.dcim.devices.filter(include=['config_context'])
[<pynautobot.models.dcim.Devices ('demo-cisco-xe') at 0x7f5c64d07b90>, <pynautobot.models.dcim.Devices ('jeff-8000v') at 0x7f5c64d1f410>, <pynautobot.models.dcim.Devices ('nxos9k') at 0x7f5c64d25210>, <pynautobot.models.dcim.Devices ('veos') at 0x7f5c64d26dd0>]
```

tsm1th:
This would close #237 if approved

Added tests to confirm context could be retrieved with the following:

```python
nautobot.dcim.devices.all(include=['config_context'])
```

### DIFF
--- a/pynautobot/core/endpoint.py
+++ b/pynautobot/core/endpoint.py
@@ -81,7 +81,7 @@ class Endpoint(object):
             ret = Record
         return ret
 
-    def all(self, api_version=None, limit=None, offset=None):
+    def all(self, *args, **kwargs):
         """Queries the 'ListView' of a given endpoint.
 
         Returns all objects from an endpoint.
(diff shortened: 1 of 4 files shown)


pynautobot/core/endpoint.py
```diff
@@ -81,7 +81,7 @@ def _lookup_ret_obj(self, name, model):
             ret = Record
         return ret
 
-    def all(self, api_version=None, limit=None, offset=None):
+    def all(self, *args, **kwargs):
```
joewesch, line 84: Hmm, I wonder if we should instead try and preserve the current signature for better backwards compatibility.

For example, if you call `nb.dcim.devices.all("1.2", 100, 0)` in the current version it is equivalent to `nb.dcim.devices.all(api_version="1.2", limit=100, offset=0)`. I don't think we need `*args` for this call so maybe just adding `**kwargs` is all we need here:
```suggestion
    def all(self, api_version=None, limit=None, offset=None, **kwargs):
```

Thoughts?

We would also need to change the call to filter:
```python
return self.filter(api_version=api_version, limit=limit, offset=offset, **kwargs)
```

tsm1th: I was back and forth on it.  I landed on including `*args` to keep similar functionality found in `filter()`:
```python
nb.dcim.devices.all("ams01-leaf-08")
```

I seriously doubted anyone was doing `nb.dcim.devices.all("1.2", 100, 0)` with the current version.  There weren't any test harmed when making this change :)

I am not against it, but I do like the simplicity of just passing `*args` and `**kwargs` to `filter()`.

joewesch (approved):

PR #257, opened 2024-11-18
Added back first_run logic that was removed in error

### LINKED ISSUE #256: Pagination issue when MAX_PAGE_SIZE is explicitly set in Nautobot
Version 2.3 introduced a bug in the pagination logic that occurs when a Nautobot instance explicitly sets a MAX_PAGE_SIZE.  

Steps to reproduce:
1) Set MAX_PAGE_SIZE lower than object count for any given object `MAX_PAGE_SIZE=1000`
2) Query for that object `prefixes = nb.ipam.prefixes.filter(status=["Active"])`
3) Observe the requests and notice the `limit` returned by API is never honored.  This causes an endless loop.
```
11:50:31.657 INFO    django.server : "GET /api/ipam/prefixes/?status=Active HTTP/1.1" 200 78502
11:50:35.685 INFO    django.server : "GET /api/ipam/prefixes/?status=Active&limit=2124&offset=50 HTTP/1.1" 200 1572898
11:51:18.839 INFO    django.server : "GET /api/ipam/prefixes/?status=Active&limit=2124&offset=1000 HTTP/1.1" 200 1581227
```
Expected requests:
```
11:55:11.655 INFO    django.server : "GET /api/ipam/prefixes/?status=Active HTTP/1.1" 200 78502
11:55:15.067 INFO    django.server : "GET /api/ipam/prefixes/?status=Active&limit=2124&offset=50 HTTP/1.1" 200 1572898
11:55:18.981 INFO    django.server : "GET /api/ipam/prefixes/?limit=1000&offset=1050&status=Active HTTP/1.1" 200 1580176
11:55:19.630 INFO    django.server : "GET /api/ipam/prefixes/?limit=1000&offset=2050&status=Active HTTP/1.1" 200 117918
```

tsm1th:
Closes #256.  This logic was previously removed in error.  The conditions that require this logic were not available in testing.  We have to account for the first query's limit not being honored by the server and force subsequent requests to use the `next` URL returned by the server.  This adds back the required logic for that scenario.

### DIFF
--- a/pynautobot/core/query.py
+++ b/pynautobot/core/query.py
@@ -327,11 +327,13 @@ class Request(object):
             req = self._make_call(add_params=add_params)
             if isinstance(req, dict) and req.get("results") is not None:
                 ret = req["results"]
+                first_run = True
                 while req["next"] and self.offset is None:
-                    if not add_params:
+                    if not add_params and first_run:
                         req = self._make_call(add_params={"limit": req["count"], "offset": len(req["results"])})
                     else:
                         req = self._make_call(url_override=req["next"])
+                    first_run = False
                     ret.extend(req["results"])
                 return ret
             else:


joewesch:
Are we able to replicate the conditions in the integration tests?

tsm1th:
I'm not sure.  Can we change the nautobot_config.py used for those tests?  We would need a way to set MAX_PAGE_SIZE and confirm it doesn't impact other tests.

tests/integration/test_ipam.py
```diff
@@ -26,3 +27,14 @@ def test_prefixes_successfully_stringify_tags(nb_client):
...
     assert isinstance(prefix.tags[0], Record)
+
+
+@pytest.mark.timeout(120)
+def test_prefixes_pagination_with_max_page_size(nb_client):
+    """Validate prefixes are returned when dataset is larger than max_page_size and not in endless loop."""
+    for i in range(0, 256):
+        nb_client.ipam.prefixes.create(
+            prefix=f"192.1.{str(i)}.0/24", namespace={"name": "Global"}, status={"name": "Active"}
+        )
```
joewesch, line 38: Wouldn't this be better as a comprehension so it's just 1 POST call?

tsm1th: of course!

PR #294, opened 2025-03-16
avoid using a re-exported stdlib urlparse

cardoe:
urlparse comes from the stdlib module urllib.parse but instead the code was relying on the re-export.

## What's Changed

Internal code style cleanup.

### DIFF
--- a/changes/294.housekeeping
+++ b/changes/294.housekeeping
@@ -1,0 +1,1 @@
+Use direct import from stdlib instead of requests re-export.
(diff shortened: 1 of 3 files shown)


joewesch:
Thanks for the contribution, @cardoe. It looks like it's just missing a [changelog fragment](https://docs.nautobot.com/projects/core/en/stable/development/core/#creating-changelog-fragments). If there is no issue number to reference, you can use this PR number (294) instead for the filename.

joewesch (approved):

PR #305, opened 2025-04-25
Adding ruff

joewesch:
# Closes: #301

## What's Changed

Replaced black, bandit, flake8 and pydocstyle with ruff.


cardoe:
@joewesch if this is good to go, I'll rebase #308 on top.

jvanderaa (approved):
Looks good generally. Only question was the part of the overrides statement on DCIM and IPAM.

joewesch:
> Only question was the part of the overrides statement on DCIM and IPAM.

Put simply, most fields should work out of the box. Some fields, however, need to be overridden from the default. The field type to require this the most is JSON types.

PR #308, opened 2025-04-28
fix: clarify behavior of update() method on Endpoint

### LINKED ISSUE #307: update() on Endpoint is documented incorrectly
For context, I'm attempting to add typing to this library so I'm looking to note discrepancies.

https://github.com/nautobot/pynautobot/blob/0eb16d295b82e492a83114d240003d3a2df489fa/pynautobot/core/endpoint.py#L307-L309

Documents that the `update()` method on an Endpoint will return either a list of `Record` or just one `Record` based on if it does a `bulk_update()` underneath or not.

But the reality of the code is https://github.com/nautobot/pynautobot/blob/0eb16d295b82e492a83114d240003d3a2df489fa/pynautobot/core/endpoint.py#L344-L345 that if a `bulk_update()` is called then that's the only time you can get a `Record` and it'll be a list of them because below https://github.com/nautobot/pynautobot/blob/0eb16d295b82e492a83114d240003d3a2df489fa/pynautobot/core/endpoint.py#L354-L356 you can only return `True` or `False`.

cardoe:
The update() method on the endpoint incorrectly documented what the return values were. The code examples didn't expose this because in the incorrect case the return was not used. Updated the documentation to reflect the behavior but also rewrote the code to narrow down the possibilities of what can be supplied so that there are narrower types allowing more type checking in the future to be added. Fixes #307

### DIFF
--- a/changes/307.fixed
+++ b/changes/307.fixed
@@ -1,1 +1,1 @@
-Fixed documentation update() method on Endpoint and added type enforcement.
+Fixed documentation for the update() method on Endpoint and added type enforcement.


joewesch (approved):
I like the use of `typing.overload`, great work.

joewesch (approved):

PR #390, opened 2026-06-02
Fix get_custom_fields and get_custom_field_choices ignoring the filters argument

### LINKED ISSUE #389: get_custom_fields() and get_custom_field_choices() silently ignore the filters argument
### Environment

- **pynautobot version:** 2.7.2 (also present on `develop`)
- **Python version:** 3.10+
- **Nautobot version:** N/A (client-side bug, version-independent)

### Summary

`App.get_custom_fields(filters=...)` and `App.get_custom_field_choices(filters=...)` accept a `filters` argument, but it is silently discarded. The methods always query the full unfiltered endpoint regardless of what the caller passes.

### Steps to Reproduce

```python
import pynautobot

nb = pynautobot.api("http://localhost:8000", token="<token>")

# Intent: only custom fields applicable to dcim.device
fields = nb.extras.get_custom_fields(filters={"content_types": "dcim.device"})
```

### Expected Behavior

The request to `/api/extras/custom-fields/` includes `content_types=dcim.device` as a query parameter, and only matching custom fields are returned.

### Observed Behavior

The `filters` argument is ignored. No filter query parameters are sent, and **all** custom fields are returned. The same applies to `get_custom_field_choices()`.

### Root Cause

In `pynautobot/core/app.py`, both methods do:

```python
filters = self.api.default_filters.copy()   # rebinds the parameter to a new dict
filters.update(filters)                      # updates a dict with itself -> no-op
```

The local assignment shadows the `filters` parameter, and `filters.update(filters)` is a no-op self-update, so the caller's value never reaches the `Request`. (Contrast with `Endpoint.filter()` / `Endpoint.get()`, which correctly do `filters = self.api.default_filters.copy(); filters.update(kwargs)`.)

### Proposed Fix

Merge the caller's filters onto a separate copy of the API defaults:

```python
request_filters = self.api.default_filters.copy()
if filters:
    request_filters.update(filters)
```

A regression test should assert that a caller-supplied filter is forwarded as a request query parameter for both methods.

jvanderaa:
# Closes: #389

## What's Changed

`App.get_custom_fields()` and `App.get_custom_field_choices()` both accepted a
`filters` argument that was silently discarded, so callers could never scope these
requests and always got the full unfiltered result set.

The cause was a self-shadowing bug. Each method rebound the `filters` parameter to a
copy of the API default filters and then called `filters.update(filters)` (a no-op
self-update), dropping the caller's value before it reached the `Request`:

```python
# before
filters = self.api.default_filters.copy()
filters.update(filters)
```

The fix merges the caller's filters onto a separate copy of the API defaults, matching
the pattern already used by `Endpoint.filter()` and `Endpoint.get()`:

```python
# after
request_filters = self.api.default_filters.copy()
if filters:
    request_filters.update(filters)
```

The API default filters (`exclude_m2m`, `include`) are preserved, and a caller's filter
of the same key takes precedence.

### Before / After

```python
nb.extras.get_custom_fields(filters={"content_types": "dcim.device"})
```

- **Before:** request sent to `/api/extras/custom-fields/` with no filter params; all custom fields returned.
- **After:** request includes `content_types=dcim.device`; only matching custom fields returned.

### Tests

Added two regression tests (`test_custom_fields_passes_filters`,
`test_custom_field_choices_passes_filters`) asserting the caller's filter is forwarded
as a request query parameter. Both fail against the old code and pass with the fix.

### Documentation

Added a `filters` usage example to both method docstrings. These render into the
auto-generated API reference page (`docs/dev/code_reference/core/app.md`), so the
documented behavior now matches what the code does.

### DIFF
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -143,14 +143,14 @@ class App:
             >>> nb.extras.get_custom_fields(filters={"content_types": "dcim.device"})
             [...]
         """
-        request_filters = self.api.default_filters.copy()
+        default_filters = self.api.default_filters.copy()
         if filters:
-            request_filters.update(filters)
+            default_filters.update(filters)
         return Request(
             base=f"{self.api.base_url}/{self.name}/custom-fields/",
             token=self.api.token,
             http_session=self.api.http_session,
-            filters=request_filters,
+            filters=default_filters,
         ).get()
 
     def get_custom_field_choices(self, filters=None):
@@ -192,14 +192,14 @@ class App:
             >>> nb.extras.get_custom_field_choices(filters={"field": "test_custom_field_2"})
             [...]
         """
-        request_filters = self.api.default_filters.copy()
+        default_filters = self.api.default_filters.copy()
         if filters:
-            request_filters.update(filters)
+            default_filters.update(filters)
         return Request(
             base=f"{self.api.base_url}/{self.name}/custom-field-choices/",
             token=self.api.token,
             http_session=self.api.http_session,
-            filters=request_filters,
+            filters=default_filters,
         ).get()
 
     def config(self):
--- a/pynautobot/core/app.py
+++ b/pynautobot/core/app.py
@@ -136,14 +136,21 @@ class App:
                     "notes_url": "http://localhost:8000/api/extras/custom-fields/5b39ba88-e5ab-4be2-89f5-5a016473b53c/notes/",
                 },
             ]
+
+            Scope the results by passing a ``filters`` dict, whose key/value
+            pairs are sent as query parameters to the endpoint:
+
+            >>> nb.extras.get_custom_fields(filters={"content_types": "dcim.device"})
+            [...]
         """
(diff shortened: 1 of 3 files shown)


pynautobot/core/app.py
```diff
@@ -136,14 +136,21 @@ def get_custom_fields(self, filters=None):
...
+
+            Scope the results by passing a ``filters`` dict, whose key/value
+            pairs are sent as query parameters to the endpoint:
+
+            >>> nb.extras.get_custom_fields(filters={"content_types": "dcim.device"})
+            [...]
         """
-        filters = self.api.default_filters.copy()
-        filters.update(filters)
+        request_filters = self.api.default_filters.copy()
```
joewesch: `request_filters` makes it sound like these came in from the request. Can you rename this to `default_filters`?

joewesch (commented):
Wow, not sure how I completely missed that.

joewesch (approved):

